PR #11, opened 2019-04-06
fix issue with parsing <float>*pi type expressions

balopat:
This fixes parsing qasm expressions like `rz(0.75*pi)` that currently fails with 

```
ValueError: could not convert string to float: '0.75*pi'
```


jvdwetering:
Thanks!

PR #13, opened 2019-07-11
load qasm bug fix

alexcowtan:
Requires ZPhase and XPhase to load Circuit from QASM file. Otherwise, throws:
`NameError: name 'ZPhase' is not defined`

If you are open to PRs in general there are some others I can submit in future.


jvdwetering:
Hey Alex,
We are very open to pull requests. Thank you for fixing this!

PR #14, opened 2019-08-02
naive cz circuit optimisation

alexcowtan:
This method takes CZ circuits and naively applies the optimisation from fig. 1, p4 of https://arxiv.org/pdf/1705.09176.pdf.

I don't know if this is where you want this method -- if you want it to go in `clifford_extract` or `streaming_extract` instead/as well then I can do that. I also know `modified_extract` might not be stable enough for new code right now.

### DIFF
--- a/pyzx/extract.py
+++ b/pyzx/extract.py
@@ -970,6 +970,41 @@ def simple_extract(g, quiet=True):
     
     return h
 
+# O(N^3)
+def max_overlap(cz_matrix):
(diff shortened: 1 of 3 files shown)


jvdwetering:
I think this is a good place to put it. At some point the modified_extract will replace the other extraction methods because it is better in every way then the previous ones (as far as I'm aware currently). Could you make this CZ optimization into an option specified by a parameter?

PR #21, opened 2020-01-28
Ignore failing test case when numpy is not installed

ABorgna:
The function `compare_tensors` is not imported in `test_graph.py` if numpy is not installed.

A test case that depends on it did not have the `skipUnless` decorator, and was causing a `NameError` when running the test suite without having numpy. There are other tests that already use this guard.


jvdwetering:
Thank you for the fix!

PR #24, opened 2020-01-30
Use an enum for vertex and edge types in the graph

ABorgna:
This PR is just a syntactic change to the graph internals.
It defines the `VertexType` and `EdgeType` enums, mantaining the same number encoding as before but allowing us to refer to them by their proper names, so functions documentation do not need to specify " (1=regular, 2=Hadamard edge)".

The main benefit is that it makes easier and less error-prone to add other node types when experimenting with variations of the ZX-calculus that define new types of nodes.

Since `IntEnum` is a subclass of `int`, the change is retrocompatible and works with all the examples. The values are transparently casted to integers when needed, but the API produces an error when setting an invalid type:
```python
>>> g.set_type(0,1)
>>> g.type(0)
<VertexType.Z: 1>
>>> g.set_type(0, 4)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    g.set_type(0, 4)
  File "[...]/pyzx/pyzx/graph/graph_s.py", line 179, in set_type
    t = VertexType(t)
  File "/usr/lib/python3.8/enum.py", line 304, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib/python3.8/enum.py", line 595, in __new__
    raise exc
  File "/usr/lib/python3.8/enum.py", line 579, in __new__
    result = cls._missing_(value)
  File "/usr/lib/python3.8/enum.py", line 608, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 4 is not a valid VertexType
```
It also lets us replace expressions like `(t1 == 1 and t2 == 2) or (t1 == 2 and t2 == 1): #types are ZX & different` by `t1 != t2 and t1.is_zx() and t2.is_zx()`.
All unit tests pass, and the benchmarks are roughly the same as in master.

As a side note, the `graph_ig` modifications are not being tested here because if we use that graph engine in the unit tests they fail even in master. I will make the PR with those fixes afterwards.

Please tell me what you think of this, and if I should change anything.


jvdwetering:
Thank you for the work. The reason we didn't do something similar to this is because we were worried it would lead to worse performance. Could you report on what kind of slowdown (if any) this incurs? Right now you have only implemented this change on the graph API but not yet on the functions in `rules` where many checks against vertex and edge type are done. I can imagine that this change could drastically slow these functions down.

Doing just a very simple test in a Jupyter notebook:
```
a = 1
%timeit a == 1
>>> 47.3 ns ± 0.95 ns per loop

from enum import IntEnum
class VertexType(IntEnum):
    normal = 1
    other = 2

a = VertexType.normal
%timeit a == VertexType.normal
>>> 176 ns ± 1.72 ns per loop
```
I.e. these checks are 3 times slower. I can also imagine that storing an enum type for every vertex and edge could require way more memory, although this is less of a concern, as the graphs we typically use aren't *that* large.

jvdwetering:
I should also add that the other graph backends like `graph_ig` haven't been updated in a very long time now, and probably will require substantial work to be made as fully functional as the default `graph_s` class. If you really wish, you can try to fix it, but I don't know whether it is really worth it.

ABorgna:
You are right about the performance hit. After replacing the integer comparisons by enums in the rules, the times went up by up to 60% for the larger benchmarks.
It is a pity, but it was a nice exercise.

I changed the enums to constants and the performance is back to normal. It helps make the rule definitions more explicit, but it doesn't add much to the original version.
We can close the PR if you think it is not necessary.

jvdwetering:
Sorry for not coming back to you on this. So you're saying that using constants doesn't cause a significant slowdown? Because if that is the case then I think having constants is indeed better then having magic 1's and 2's hanging around everywhere.

ABorgna:
I updated the changes to the latest version and reran benchmarks.
There is no noticeable change in the timings, all the differences seem to be just noise.

Full benchmark results:
[pyzx-benchmark-constants.txt](https://github.com/Quantomatic/pyzx/files/4475510/pyzx-benchmark-constants.txt)
[pyzx-benchmark-base.txt](https://github.com/Quantomatic/pyzx/files/4475511/pyzx-benchmark-base.txt)

PR #35, opened 2020-04-23
Bialgebra rule fix

y-richie-y:
Currently the bialgebra rule is matched without checking node phases.
This PR fixes both `match_bialg` and `match_bialg` by making `match_bialg` invoke 
`match_bialg_parallel` with `num=1`.

bialgebra_rule_issue

Credit to @sg495 for providing an illustration as to why the optimisation fails with non-zero phases.

### DIFF
--- a/pyzx/rules.py
+++ b/pyzx/rules.py
@@ -91,7 +91,8 @@ def match_bialg_parallel(g, matchf=None, num=-1):
         v1t = types[v1]
         v0p = phases[v0]
         v1p = phases[v1]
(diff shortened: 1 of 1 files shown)


jvdwetering:
Great! This piece of code was one of the first things we wrote in PyZX and has never been suitably updated. I've recently added typing hints to large parts of the PyZX code base. Could you please update the PR with these annotations (just look at the master branch).

y-richie-y:
Thanks! Not an expert on Git, but is it possible that the PR is only showing changes based on my fork, and that you'll be able to merge without conflicts, keeping your hints?

PR #66, opened 2021-04-26
Added annealing and genetic algorithms for local search

rkruegs123:


pyzx/__init__.py
```diff
@@ -26,6 +26,8 @@
 from .extract import *
 from .io import *
 from .tensor import *
+from .simulated_annealing import *
+from .genetic import *
```
jvdwetering: I don't think we should import these into global scope

pyzx/__init__.py
```diff
@@ -43,7 +45,9 @@
 from . import simulate
 from . import editor
 from . import routing
-from . import sparsify
+from . import scores
+from . import simulated_annealing
+from . import genetic
 from .routing.parity_maps import CNOT_tracker
```
jvdwetering, line 49: Could we maybe put all of these under a single module, that might be a little less messy.

pyzx/congruences.py
```diff
@@ -0,0 +1,153 @@
+import numpy as np
```
jvdwetering: Missing the standard header on top of the file. Also should have a top level documentation comment.

jvdwetering: Also, there is no mypy annotation here, which I guess is fine for now, but I would like pyzx to have annotations at least for all the function signatures. All the functions are also missing documentation strings.

pyzx/genetic.py
```diff
@@ -0,0 +1,156 @@
+from tqdm import tqdm
```
jvdwetering: Missing the standard header on top of the file. Also should have a top level documentation comment. Also, there are no documentation strings for all the classes.

pyzx/simplify.py
```diff
@@ -356,3 +356,98 @@ def clifford_iter(g: BaseGraph[VT,ET]) -> Iterator[Tuple[BaseGraph[VT,ET],str]]:
     yield from pivot_iter(g)
     yield from id_iter(g)
     yield from spider_iter(g)
+
+
+def is_graph_like(g):
+    # checks that all spiders are Z-spiders
```
jvdwetering, line 364: Should be a documentation string """using these quotes""" instead of a regular comment.

pyzx/simplify.py
```diff
@@ -356,3 +356,98 @@ def clifford_iter(g: BaseGraph[VT,ET]) -> Iterator[Tuple[BaseGraph[VT,ET],str]]:
...
+    for z in zs:
+        b_neighbors = [n for n in g.neighbors(z) if g.type(n) == VertexType.BOUNDARY]
+        if len(b_neighbors) > 1:
+            return False
+
+    return True
+
+
+def to_graph_like(g):
+    # turn all red spiders into green spiders
```
jvdwetering, line 404: Same as above

pyzx/simulated_annealing.py
```diff
@@ -0,0 +1,60 @@
+from tqdm import tqdm
```
jvdwetering: Missing the standard header on top of the file. Also should have a top level documentation comment. Missing documentation strings for all the functions.

jvdwetering (commented):
If you could add a bit of documentation to all the methods you've added that would be great. Mypy annotations for at least the function signatures would also be welcome, but I understand it is a bit of a pain. I think the three files you've added could also be part of the same submodule as to not further pollute the (already crowded) namespace. Also, please add a license statement to the top of each file.

rkruegs123:
Sounds good. I've made these changes aside from the typing annotations. Happy to do this as well, but will take a bit longer.

akissinger:
I think typing is not necessary for now. I'm eager to start doing some experiments with your code.

PR #67, opened 2021-05-05
Add lookahead extraction

VladMoldoveanu:
The lookahead extraction has a main function called `lookahead_extract_base` and 3 predefined functions that offer good parametrizations for the base one: `lookahead_fast`, `lookahead_extract`, `lookahead_full`. 

Changes to the existing code include splitting the existing extraction into smaller pieces and reusing them where needed, a function for identical copies for `GraphS`, and a function in `Circuit` that computes the depth of a circuit.

An additional change that I found very useful in testing, but is not needed in the extraction, is a function in `Circuit` that returns a dictionary of the different parameters for a circuit. It is similar to `Circuit.stats()`, but the dictionary allows for different values to be compared between circuits, while the string output only allowed printing.

### DIFF
--- a/pyzx/extract.py
+++ b/pyzx/extract.py
@@ -382,6 +382,191 @@ def filter_duplicate_cnots(cnots: List[CNOT]) -> List[CNOT]:
     c = basic_optimization(c,do_swaps=False)
     return c.gates # type: ignore
 
+
+def remove_extra_cnots(cnots_to_apply: List[CNOT], mat: Mat2) -> List[CNOT]:
+    """Remove redundant CNOTs and return a list of only the necessary ones"""
+    cnots = cnots_to_apply.copy()
+    m2 = mat.copy()
+    for cnot in cnots:
+        m2.row_add(cnot.target, cnot.control)
+    extractable = set()
+    for i, row in enumerate(m2.data):
+        if sum(row) == 1:
+            extractable.add(i)
+    # We now know which vertices are extractable, and hence the CNOTs on qubits that do not involve
+    # these vertices aren't necessary.
+    # So first, we get rid of all the CNOTs that happen in the Gaussian elimination after
+    # all the extractable vertices have become extractable
+    m2 = mat.copy()
+    for count, cnot in enumerate(cnots):
+        if sum(1 for row in m2.data if sum(row) == 1) == len(extractable):  # extractable rows equal to maximum
+            cnots = cnots[:count]  # So we do not need the remainder of the CNOTs
+            break
+        m2.row_add(cnot.target, cnot.control)
+    # We now recalculate which vertices were extractable, because the deleted cnots
+    # might have acted to swap this vertex around some.
+    extractable = set()
+    for i, row in enumerate(m2.data):
+        if sum(row) == 1:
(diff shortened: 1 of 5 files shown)


pyzx/extract.py
```diff
@@ -222,6 +229,115 @@ def greedy_reduction(m: Mat2) -> Optional[List[Tuple[int,int]]]:
         indices.remove(control)
     return result
 
+
+def flat_indices(m: Mat2, indices: list[int]) -> Tuple[List[Tuple[int, int]], int]:
```
jvdwetering: Please add a docstring to the function

pyzx/extract.py
```diff
@@ -222,6 +229,115 @@ def greedy_reduction(m: Mat2) -> Optional[List[Tuple[int,int]]]:
...
+        next_indices.append(target)
+        if len(indices) <= 1:
+            if len(indices) == 1:
+                next_indices.append(indices[0])
+            indices = next_indices
+            next_indices = []
+    return result, indices[0]
+
+
+def greedy_reduction2(m: Mat2) -> Optional[List[Tuple[int, int]]]:
```
jvdwetering: Perhaps have a more descriptive name than just adding a 2 at the end? Or at least make more clear in the docstring how it differs.

pyzx/extract.py
```diff
@@ -222,6 +229,115 @@ def greedy_reduction(m: Mat2) -> Optional[List[Tuple[int,int]]]:
...
+                combs2[(*index, k)] = row
+                iterations += 1
+            if iterations > 100000:
+                return None
+        if not combs2:
+            return None
+        combs = combs2
+
+
+def greedy_reduction3(m: Mat2) -> Optional[List[Tuple[int, int]]]:
```
jvdwetering: Same as above

pyzx/extract.py
```diff
@@ -266,11 +382,197 @@ def filter_duplicate_cnots(cnots: List[CNOT]) -> List[CNOT]:
...
+
+    for cnot in cnots:
+        c.add_gate(cnot)
+    for h in hads:
+        c.add_gate("HAD", h)
+
+    return len(good_verts)
+
+
+def clean_frontier(g: BaseGraph[VT, ET], c: Circuit, frontier: List[VT],
```
jvdwetering, line 490: docstring

pyzx/extract.py
```diff
@@ -266,11 +382,197 @@ def filter_duplicate_cnots(cnots: List[CNOT]) -> List[CNOT]:
...
+
+    for i in range(g.qubit_count()):
+        for j in range(i + 1, g.qubit_count()):
+            if cz_mat.data[i][j] == 1:
+                c.add_gate("CZ", i, j)
+
+    return czs_saved
+
+
+def neighbors_of_frontier(g: BaseGraph[VT, ET], frontier: List[VT]) -> Set[VT]:
```
jvdwetering, line 538: docstring

jvdwetering (commented):
Thanks for the pull request! It looks great. I just put some comments about missing docstrings for functions.

jvdwetering:
Thanks for the work and the fixes!

PR #73, opened 2021-09-14
Add support for ground generators in the graphs

ABorgna:
Hi, this is a proposal for adding support for the ground generator (⏚) of the ZX⏚ calculus, to be able to represent partial traces and classical operations. I used this as a basis to implement the optimization routines in my hybrid optimization paper ([arXiv:2109.06071](https://arxiv.org/abs/2109.06071)) which I'll submit in a subsequent PR.

## Implementation

Instead of adding a separate kind of node, the spiders have a new boolean flag signaling if they have a single connected ⏚ (all diagrams can be normalized to this form). This is added directly to `BaseGraph` akin to the API for `row` and `qubit`. It defines the methods [`is_ground`, `set_ground`] for specific vertices and [`grounds`, `is_hybrid`] for the whole graph. They are implemented on `GraphS`.

The rewriting rules in `rules` had to be modified to check for the flag, as they should consider ⏚-spiders as non-Clifford for the rewriting steps regardless of their phase.

The `drawing` module now draws the ground symbols on the nodes, and is able to optionally show vdata fields.

image

There is a simple extension of the `cliffordT` circuit generator, `cliffordTmeas`, which can insert measurement gates that ignore their classical result and just collapse the qubit state.

Finally, the `Circuit`s now support bit registers, with their own id separate from the qubits to better match the circuit description languages. These are all transformed into _qubits_ when transformed into a graph, as there is no difference between them in the diagrams. The `Gate.to_graph` methods where reworked a bit to clean them up and to support bit registers.

## Backwards compatibility

The only change in the API is that the `BaseGraph` class now has some ground-related abstract methods, which need to be implemented on new instances.

Algorithms implemented for pure circuit do not need to be changed, although it may be a good idea to check for `g.is_hybrid()` and raise an exception if it is not supported.

The benchmarks do not show any significant difference in performance on the existing tests.

## Alternatives

This could have been implemented as vdata labels on the vertices, but it would not standardize the way of representing grounds and may lead to algorithms mistakenly ignoring them (for example, unknowingly using `compare_tensors` on two hybrid graphs).
Our way let's us include an explicit `Graph.is_hybrid` method.

Another option would be to use a graph wrapper inheriting from `BaseGraph`. It'd be similar to this solution but making an algorithm that works both in pure-quantum and hybrid graphs would be more tedious and inefficient.

## Things not in this PR

- Optimization routines from the paper. I left them out to reduce the noise on an eventual design change.
I have them implemented in a separate submodule, which I'll submit once this PR is merged.

- Classicalization routines. The paper describes an algorithm for detecting wires which carry classical data.
This is implemented using a graph wrapper that adds this information to the edges, so it too would be too noisy for this PR.

- Qasm circuit parsing. The current implementation of the parser is not updated for OPENQASM 3,
so I think it will be better to rewrite it all at once. (`Circuit` *does* support bit registers).

- Interacting with gadget phase teleportation. The ground nodes can be interpreted as gadgets with an absorbing element '⏚' as it's phase. It shouldn't be too hard to add this later, but it was not the main goal of the change.

- Support for ⏚s in the js editor.

- `compare_tensor`, but with CPTP maps.

- Documentation. I'll add it along with the optimization routines.

### DIFF
--- a/pyzx/circuit/__init__.py
+++ b/pyzx/circuit/__init__.py
@@ -19,7 +19,7 @@ from typing import List, Union, Optional, Iterator, Dict
 
 import numpy as np
 
-from .gates import Gate, gate_types, ZPhase, XPhase, CZ, CX, CNOT, HAD, SWAP, CCZ, Tofolli
+from .gates import Gate, gate_types, ZPhase, XPhase, CZ, CX, CNOT, HAD, SWAP, CCZ, Tofolli, Measurement
 
 from ..graph.base import BaseGraph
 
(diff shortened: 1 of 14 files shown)


jvdwetering:
Sorry this took so long, somehow I believed I had already accepted this PR.

ABorgna:
Great! I'll make another the PR with the optimization routines once I have the time.

PR #80, opened 2021-11-15
Added a function to_tensor to convert a ZX-diagram to a tensor network.

paultirlisan:
As part of my third-year project, the first task to is write a function that can translate a ZX-diagram (in this case a **BaseGraph**) to a **TensorNetwork**; the library used for tensors is [Quimb](https://quimb.readthedocs.io/en/latest/).

Three unit tests have been written for this function:
- one that tests a simple diagram with two boundary vertices connected to each other, i.e. the identity;
- one that tests a diagram consisting of two boundary vertices connected by a blue edge, i.e. the Hadamard gate;
- a graph with two inputs and one output representing the XOR gate.


jvdwetering:
Hi, thanks for the work! I've looked at the code, and I don't know much about quimb, so I can't really check whether this is correct. Could you maybe write some tests or something to include so that I can verify that it indeed works as desired (I'm assuming you have tested this yourself with some simple graphs, so that shouldn't take too much time I hope).
The CI is also showing some mypy errors, but I wouldn't worry too much about that.

paultirlisan:
Hi! I had a discussion with Aleks and we agreed on changing the implementation for that function since it was incorrect. I will soon commit the fixed version, together with three unit tests.

tests/test_mapping.py
```diff
@@ -0,0 +1,98 @@
...
+except ImportError:
+    np = None
+
+try:
+    import quimb as qu
+    import quimb.tensor as qtn
+except ImportError:
+    qu = None
+
+from pyzx.graph import Graph
```
jvdwetering: In this line you should do `from ..graph import Graph` instead of `pyzx.graph`, as that would make it look at installed packages instead of in the directory above it.

pyzx/__init__.py
```diff
@@ -20,6 +20,7 @@
 from .circuit import Circuit, gates, id
 from .linalg import Mat2
 from .utils import settings, VertexType, EdgeType
+from .mapping import to_tensor
```
jvdwetering: I'm not sure `mapping` is a very descriptive name. Maybe just call it `quimb` or `tensor_network`?

tests/test_mapping.py
```diff
@@ -0,0 +1,98 @@
...
+        
+        tn = to_tensor(g)
+        self.assertTrue(abs((tn & qtn.Tensor(data = [1, 0], inds = ("0",)) 
+                            & qtn.Tensor(data = [1 / np.sqrt(2), 1 / np.sqrt(2)], inds = ("1",)))
+                        .contract(output_inds = ()) - 1) < 1e-9)
+        self.assertTrue(abs((tn & qtn.Tensor(data = [0, 1], inds = ("0",)) 
+                            & qtn.Tensor(data = [1 / np.sqrt(2), -1 / np.sqrt(2)], inds = ("1",)))
+                        .contract(output_inds = ()) - 1) < 1e-9)
+    
+    def test_xor_tensor(self):
```
jvdwetering: I would suggest adding a test (or changing this test) to include some phases on some of the spiders, to test whether you have implemented these phases correctly.

jvdwetering (commented):
Added some comments

PR #97, opened 2022-08-30
Support representing symbolic graphs

y-richie-y:
This is based on the work of Matthew Sutcliffe. For now this only allows the creation and drawing of symbolic graphs. ~~By using sympy's pretty print features, this `symbolize` function should be more robust.~~ Now the code just converts the symbolic expression to string, which should still look fine.

```python
import pyzx as zx
from sympy import symbols

circ = zx.Circuit(3)
syms = symbols('α β γ')
for i, sym in enumerate(syms):
    circ.add_gate('XPhase', i, -(i + 1) * sym ** (i + 2) + 1)
zx.draw(circ.to_graph())
```

image

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -32,6 +32,3 @@ ignore_missing_imports = True
 
 [mypy-IPython.*]
 ignore_missing_imports = True
-
-[mypy-sympy.*]
-ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -32,3 +32,6 @@ ignore_missing_imports = True
 
 [mypy-IPython.*]
 ignore_missing_imports = True
+
+[mypy-sympy.*]
+ignore_missing_imports = True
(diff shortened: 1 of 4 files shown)


jvdwetering:
Hi Richie, thanks for the work. It looks like this PR makes sympy a dependency of PyZX. Isn't that a really big package? It might be good to make this a conditional import.

jvdwetering:
I'm also surprised this isn't making mypy complain, as it doesn't look like you've updated the `FractionLike` union type.

y-richie-y:
I guess it is because I had to disable sympy related typing analysis, since sympy doesn't have mypy stubs.

y-richie-y:
> Hi Richie, thanks for the work. It looks like this PR makes sympy a dependency of PyZX. Isn't that a really big package? It might be good to make this a conditional import.

I'm actually inclined to remove mentions of sympy, and opt for a try-except type approach (e.g. try converting to Fraction). This would make things a lot lighter.

Does that sound ok?

jvdwetering:
I think that is fine for now. The try-excepts might make it a bit slower though. But since PyZX is not built for speed, that is probably fine.

y-richie-y:
https://stackoverflow.com/questions/2522005/cost-of-exception-handlers-in-python

worth mentioning that try-excepts are "not that bad" - the try itself does not slow things down, and the case where an exception is raised is of "the same order of magnitude as an if-else".

PR #99, opened 2022-09-02
Add more symbolic functionality

y-richie-y:
From my experimentation this enables `full_reduce` on symbolic diagrams. Here is an example on a DisCoCat diagram:

image
image

### DIFF
--- a/pyzx/circuit/gates.py
+++ b/pyzx/circuit/gates.py
@@ -563,10 +563,17 @@ class CRZ(Gate):
         self.phase = phase
 
     def to_basic_gates(self):
-        return [ZPhase(self.target,Fraction(self.phase/2)),
-                CNOT(self.control,self.target),
-                ZPhase(self.target,Fraction(-self.phase/2)%2),
-                CNOT(self.control,self.target)]
+        phase1 = self.phase / 2
+        phase2 = -self.phase / 2
+        try:
+            phase1 = Fraction(phase1) % 2
+            phase2 = Fraction(phase2) % 2
+        except Exception:
+            pass
+        return [ZPhase(self.target, phase1),
+                CNOT(self.control, self.target),
+                ZPhase(self.target, phase2),
+                CNOT(self.control, self.target)]
 
 
     def to_graph(self, g, q_mapper, c_mapper):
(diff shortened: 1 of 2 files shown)


y-richie-y:
Hm is this a bug I just introduced?
image
image

I don't see how phases `START__s_2` and `test__s.r@s_0/2` can be made adjacent.

jvdwetering:
Yeah it looks like something weird is going on. Can you try the same diagram, but replacing each parameter by some non-canonical non-Clifford phase, to see if the same thing happens?

jvdwetering:
Then you can also directly evaluate the matrices to see if it is actually a bug

y-richie-y:
still happens, seems unrelated to the PR
image
image


They appear to be equal!

```
array([[ 0.50073846+0.00004646j,  0.50073846+0.00004646j,
         0.49704253+0.0470077j ,  0.49704253+0.0470077j ,
         0.49827673+0.03132563j,  0.49827673+0.03132563j,
        -0.4945808 -0.07828686j, -0.4945808 -0.07828686j],
       [ 0.49926045-0.00002324j, -0.49926045+0.00002324j,
         0.49852053+0.0470774j , -0.49852053-0.0470774j ,
         0.50073846+0.00004646j, -0.50073846-0.00004646j,
        -0.49704253-0.0470077j ,  0.49704253+0.0470077j ]])
```

y-richie-y:
Do you understand this? @jvdwetering

jvdwetering:
Are you sure this is not a case of spiders appearing to be on top of each other in the drawing? Are they actually connected?

y-richie-y:
I'm dragging them, they appear to be connected. Let me try to produce a minimal example.

jvdwetering:
I think this must be the result of a boundary pivot on the spider connected to the second input and the hub part of the phase gadget.

y-richie-y:
Just to confirm, so we think there is no bug right?

jvdwetering:
Well, if the matrix is preserved, then it is a correct rewrite.

y-richie-y:
Ok. The first change I believe is correct, as it appears to be checking if the spider is Clifford.
```
if phases[v] != 0 and hasattr(phases[v], 'denominator') and phases[v].denominator > 2 and len(list(g.neighbors(v)))==1:
```

In the second change, it is looking for phase gadgets, so is it the case that we want to **include** non-clifford phases, and also is it something that would work anyway if we didn't match it?

y-richie-y:
@jvdwetering do you think this is good to merge?

jvdwetering:
From what I saw in your changes this all looks fine. But aren't there more rewrite rules you need to change to make them compatible with symbolic angles?

y-richie-y:
I'll have another look tomorrow, but full_reduce has been working for me in the wild.

PR #107, opened 2023-01-23
Move cnot_mapper to routing and add functionality

aborgna-q:
This PR continues with the upstream of some of the changes implemented by @Aerylia in [CQCL/pyzx](https://github.com/CQCL/pyzx/tree/steiner_decomp) while working for Cambridge Quantum.

This one updates the architecture-constrained CNOT mapping routines, moving them from a script to the `routing` module, and adding tests all around.

We found a failing edge case in the existing implementation of `routing.steiner` that we weren't able to fix. There's a (skipped) test case at the end of `test_cnot_mapper` that shows it.

These changes are required to implement the phase-polynomial synthesis procedure from Arianne's [paper](https://arxiv.org/abs/2004.06052), which I will send in a following PR.

### Changes

- Extract the cnot_mapper procedure from the script, and moves it to `routing.cnot_mapper` so it can be used by the rest of the code, and moves the script to `scripts.circuit_router`

- Add more fitness functions for the genetic cnot mapping algorithm.

- Move the cnot generator script from `routing.parity_maps` to its own script, `scripts.cnot_generator`

- Add recursive implementation in `routing.steiner`, and a partial column reduction function

- Add tests for cnot_mapper and `routing.steiner`

- Add a Particle Swarm Optimization (PSO) to machine_learning, used by one of the new cnot_mapping modes


jvdwetering:
Hi Augustin, great! The CI is complaining because networkx is not listed as a standard dependency of pyzx. I believe networkx is quite a large install, so I would like to keep it that way. Could you add some kind of conditional skip around the test so that it is ignored if networkx is not installed?

aborgna-q:
networkx was being used in a couple Architecture generators, but I think it can be replaced by pyzx's graph. I'll update that in a moment.

PR #108, opened 2023-01-23
Add routing methods for phase polynomials

aborgna-q:
Implements the phase polynomial routing method described in the paper [Architecture-aware synthesis of phase polynomials for NISQ devices](https://arxiv.org/abs/2004.06052) by Arianne Meijer-van de Griend and Ross Duncan.

Includes a set of circuits to run tests on, and a script to generate random phase polynomial circuits.

This is the last PR upstreaming of some of the changes implemented by @Aerylia in [CQCL/pyzx](https://github.com/CQCL/pyzx/tree/steiner_decomp) while working for Cambridge Quantum.

PR #109, opened 2023-01-27
Add some documentation for the phase polynomial routing

aborgna-q:
This PR adds more docstrings to the changes from #108.
The root module `pyzx.routing` now re-exports some common operations, which are also shown in the API reference section of the docs. I also added a short usage example there.

There are some small changes to the code like moving the circuit generation routines to `generate`, and adding an explicit `Parity` class instead of using strings of "0"s and "1"s. There should be no functionality changes.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -108,6 +108,20 @@ Below is listed the content of ``optimize.py``.
    :undoc-members:
 
 
+.. _routing:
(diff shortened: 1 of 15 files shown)


jvdwetering:
Very nice, thanks for all the work Augustín!

PR #110, opened 2023-02-16
Fix `to_quimb_tensor()`

y-richie-y:
Support phased spiders with more than 1 leg, fulfilling the condition stated in https://github.com/Quantomatic/pyzx/blob/e93a0829ef467bd6c9887332b4899e0f7c1ddc80/pyzx/quimb.py#L50

Example code
```python
from pyzx import Circuit
from pyzx.circuit.gates import Z, NOT
from pyzx.quimb import to_quimb_tensor

c = Circuit(1)
c.add_gate(Z(0))
c.add_gate(NOT(0))

g = c.to_graph()
inds = [str(i) for i in [*g.inputs(), *g.outputs()]]
print(to_quimb_tensor(g).contract(output_inds=inds).data)
```

### DIFF
--- a/pyzx/quimb.py
+++ b/pyzx/quimb.py
@@ -27,7 +27,7 @@ except ImportError:
 
 from .utils import EdgeType, VertexType
 from .graph.base import BaseGraph
-from .simplify import to_gh
+from .simplify import to_gh, unfuse_phase_spiders
 
 def to_quimb_tensor(g: BaseGraph) -> 'qtn.TensorNetwork':
     """Converts tensor network representing the given :func:`pyzx.graph.Graph`.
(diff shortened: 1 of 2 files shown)


y-richie-y:
`to_gh()` acts in-place, and the graph already gets copied within `to_quimb_tensor()`. Would you still prefer me to change it to copy it within `unfuse_phase_spiders()`?

jvdwetering:
Oh I missed that. Never mind then.

tlaakkonen:
I'm not sure this fix is necessary - the condition in the comment doesn't have anything to do with what kind of spiders are allowed to have phases. The phase tensors are *extra* spiders added to represent the phases (essentially doing exactly the unfusing you are doing). The original spiders are all converted into hyper-indices, no tensors are ever created for individual spiders. This rewrite just adds an unnecessary identity tensor that quimb will remove anyway when you do a diagonal_reduce simplification pass.

y-richie-y:
Yeah not sure why it didn't work for me in the first place. We can roll back this commit.

Issue #62, opened 2021-02-24
type annotations for numpy objects


jvdwetering:
NumPy 1.20 added type annotations, which broke mypy. I've added dummy `type: ignore`s in the relevant places, but at some point appropriate types for numpy have to be added.

PR #118, opened 2023-06-28
Fix mypy complaining about incompatible type for Z2 = Literal[0,1].

dlyongemallo:
Partially addresses #62.

### DIFF
--- a/pyzx/linalg.py
+++ b/pyzx/linalg.py
@@ -15,10 +15,8 @@
 # limitations under the License.
 
 import math
-from typing import Union, Any, Tuple, List, Optional, Set, Dict
+from typing import Union, Any, Tuple, List, Optional, Dict, cast
 from typing_extensions import Literal
-
-import numpy as np
 
 from .circuit.gates import CNOT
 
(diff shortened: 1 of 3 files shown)


jvdwetering:
Thanks for the fixes, but #62 refers specifically to type annotations for numpy objects, while your commit deals with some type:ignore's in other libraries. Which is still welcome of course!

dlyongemallo:
You're right, I should've removed that comment or made it more clear. This PR started out by reversing commit https://github.com/Quantomatic/pyzx/commit/d7db04aa8cd1e907b9d577f00a3813a4cdcaeed1 and fixing each line which had `# type: ignore`, but I wasn't able to repro the mypy errors which were actually due to numpy. It seems that the problem has actualy been fixed in mypy prior to version 0.990, which you're using.

Also, according to https://github.com/python/mypy/issues/1424, the problems with `hasattr` are fixed in 0.991, so if/when you upgrade to that version, you can remove the `# type: ignore`s from the lines with `hasattr` in the pyzx code as well.

PR #119, opened 2023-06-30
Remove a number of unnecessary "# type: ignores".

dlyongemallo:
These lines no longer cause a problem in mypy>=0.990. Also add mypy>=0.990 to requirements.txt.


jvdwetering:
Very nice. Thanks for checking.

PR #126, opened 2023-08-01
serialize and deserialize GraphDiff

RazinShaikh:


jvdwetering:
Looks like there are still some type issues. Could you fix those?

RazinShaikh:
It should be fine now

PR #130, opened 2023-08-21
Add iterators for `full_reduce`, `interior_clifford_simp` and `pivot_gadget_simp`

y-richie-y:


jvdwetering:
I don't think the iter for full_reduce is in here?

y-richie-y:
It is!
image

jvdwetering:
Oh now I see it, I guess I didn't scroll far enough in the diff :P

PR #135, opened 2023-08-24
Fix #134: incorrect optimization for qiskit circuit

### LINKED ISSUE #134: Incorrect optimization for Qiskit circuit?
Optimizing a simple Qiskit circuit seems to yield a circuit implementing a different unitary.

Here is a minimal example:

```python
import pyzx as zx
from qiskit.circuit import QuantumCircuit

# version
print('Version:', zx.__version__)

# create simple circuit
circuit_qiskit = QuantumCircuit(1)
circuit_qiskit.t(0)
circuit_qiskit.t(0)
circuit_qiskit.t(0)
circuit_qiskit.t(0)
circuit_qiskit.t(0)
circuit = zx.Circuit.from_qasm(circuit_qiskit.qasm())

# optimize
basic = circuit.to_basic_gates()
optimized = zx.optimize.basic_optimization(basic)
optimized_qiskit = QuantumCircuit.from_qasm_str(optimized.to_qasm())

# draw
print('Before optimization:')
print(circuit_qiskit)
# zx.draw(circuit)
print('After optimization:')
# zx.draw(optimized)
print(optimized_qiskit)
```

Here is the output:
```
Version: 0.7.3
Before optimization:
   ┌───┐┌───┐┌───┐┌───┐┌───┐
q: ┤ T ├┤ T ├┤ T ├┤ T ├┤ T ├
   └───┘└───┘└───┘└───┘└───┘
After optimization:
   ┌───┐
q: ┤ T ├
   └───┘
```

Am I doing something wrong?

bichselb:
This should fix https://github.com/Quantomatic/pyzx/issues/134.

### DIFF
--- a/pyzx/optimize.py
+++ b/pyzx/optimize.py
@@ -464,6 +464,7 @@ class Optimizer(object):
                 if self.availty[t] == 2: # If previous gate was of X-type
                     self.availty[t] = 1  # We reset the available gates on this qubit
                     self.available[t] = list()
+                g = ZPhase(t, g.phase)  # Avoid subclasses of ZPhase with inconsistent phase
                 self.add_gate(t, g)
         elif g.name == 'CZ':
             t1, t2 = g.control, g.target


dlyongemallo:
I noticed that this change has a side-effect on an edge case. Suppose the input is this:
```
   ┌───┐
q: ┤ T ├
   └───┘
```
Then, after applying optimisation, the previous output would've been no change:
```
   ┌───┐
q: ┤ T ├
   └───┘
```
However, with this change, the output now becomes:
```
   ┌─────────┐
q: ┤ Rz(π/4) ├
   └─────────┘
```
It's logically the same, but it's a change nevertheless. Is this desirable?

jvdwetering:
I don't think that should lead to any problems. It is semantically still the same circuit.

dlyongemallo:
I agree. I just thought I'd note it in case someone wanted a noop for an already optimised input, or if some tool which takes in qasm input treats `t` and `rz(0.25*pi)` differently for some reason.

Issue #116, opened 2023-06-01
Support for OpenQASM 3.0


fvirdia:
Hi there,

I am trying to use PyZX to attempt reducing the footprint of some circuits I designed in Qiskit.
When trying to run `pyzx.qasm` on the QASM source generated by Qiskit, I was getting an error on `import 'stdgates.inc'`, which I understand is a library of "standard" gates used by Qiskit.

I tried replacing the line with the content of [stdgates.inc](https://github.com/openqasm/openqasm/blob/main/examples/stdgates.inc), however this results in `TypeError: Arguments for custom gates are currently not supported: p(λ) a { ctrl @ gphase(λ) a; }` on the very first "phase" gate in the file. 

I could try downgrading to a previous Qiskit, however I was wondering if there is any way to support this out of the box, since anyway it seems OpenQASM 2.0 is considered deprecated.

dlyongemallo:
Hi, @fvirdia,

Is there anything else you need? PR #151 was just merged which adds support for the OpenQASM 3 header format and the `p` gate.

dlyongemallo:
PR #156 adds supports for OpenQASM 3 format. I tried some exports from qiskit and they seem to import just fine, so unless there are requests for additional qasm gates not covered by that PR, this issue can be closed.

dlyongemallo:
I think this issue can be closed now.

Issue #103, opened 2022-12-12
CRZ gate decomposition error


anitalu724:
Hi, according to `pyzx/circuit/gates.py`, your decomposition and tensor of crz(pi/2) is shown below:
![Screen Shot 2022-12-12 at 5 31 48 PM](https://user-images.githubusercontent.com/42861761/207010771-907a6060-3df2-40bb-8dad-bd0805ca0011.png)

However, the correct form of crz(pi/2) should be like this:
![Screen Shot 2022-12-12 at 5 37 20 PM](https://user-images.githubusercontent.com/42861761/207011930-32d17276-d9b0-4741-916d-ce4fb13da729.png)

Refer to this transformation:
![318993810_684855746624910_6597546469576874782_n (1)](https://user-images.githubusercontent.com/42861761/207012131-c19d36e9-e613-4ebd-9561-c5d967261861.png)

May I submit a PR to modify this bug?

jvdwetering:
So a problem here is that different libraries and people use different standards for this gate. I believe I based the current definition on the qasm spec, but I can check

jvdwetering:
Yes QASM3.0 seems to use your definition. Which is annoying because now the specs aren't compatible apparently?
I don't really know how to deal with this now.

dlyongemallo:
PR #156 specifically adds a test to document the behaviour of `rz` and `crz`. I think this issue can be closed.

dlyongemallo:
This issue can be closed.

Issue #102, opened 2022-11-27
Unitary matrix seems to change after simplifications/optimizations from pyZX


gprs1809:
**What am I trying?** 

I am trying to use circuits from Qiskit, converting them to pyZX circuits, applying pyZX optimizations/simplifications and then converting the resultant circuit to Qiskit circuit. I see that the unitary matrix from original circuit and resultant circuit changes. I would like to check if there is a bug in pyZX or if there is a different way of applying optimizations or if there is any issue with my implementation.

**My implementation**

An example circuit created using Qiskit:
`
 qc=QuantumCircuit(4)
 qc.ccx(2,1,0)
 qc.ccz(0,1,2)
 qc.h(1)
 qc.ccx(1,2,3)
 qc.t(1)
 qc.ccz(0,1,2)
 qc.h(1)
 qc.t(0)
 qc.ccz(2,1,0)
 qc.s(1)
 qc.ccx(2,1,0)
 qc.crz(0.2*np.pi,0,1)
 qc.rz(0.8*np.pi,1)
 qc.cry(0.4*np.pi,2,1)
 qc.crx(0.02*np.pi,2,0)

`
Getting unitary matrix from above, applying pyZX optimizations followed by conversion to qiskit circuit and then, getting unitary matrix again for the resultant circuit:

`
 simulator = Aer.get_backend('aer_simulator')
 qc1 = transpile(qc, simulator,optimization_level=0,basis_gates=['u3','cx'])

 qc1.save_unitary()
 result = simulator.run(qc1).result()
 mat1 = np.asarray(result.get_unitary(qc1))

 g=zx.qasm(qc1.qasm())
 g1 = g.to_graph()
 zx.simplify.full_reduce(g1)
 g1 = zx.extract_circuit(g1).to_basic_gates()

 n=g1.qubits
 qc2=QuantumCircuit(n)
 for g in g1.gates:
     print(g)
     if g.name == 'SWAP':
         qc2.swap(g.control,g.target)
     if g.name == 'HAD':
         qc2.h(g.target)
     if g.name == 'ZPhase':
         qc2.rz(g.phase*np.pi,g.target)
     if g.name == 'CZ':
         qc2.cz(g.control, g.target)
     if g.name == 'CNOT':
         qc2.cx(g.control, g.target)
     if g.name == 'S':
         qc2.rz(g.phase*np.pi, g.target)
     if g.name == 'T':
         qc2.rz(g.phase*np.pi, g.target)
[...]

PR #147, opened 2023-09-05
Add unittest to document existing behaviour of qasm2 rz and crz gates in pyzx.

dlyongemallo:
The OpenQASM 2 spec defines the following:
```
gate rz(phi) a { u1(phi) a; }
gate crz(lambda) a,b
{
    u1(lambda/2) b;
    cx a,b;
    u1(-lambda/2) b;
    cx a,b;
}
gate cu1(lambda) a,b
{
    u1(lambda/2) a;
    cx a,b;
    u1(-lambda/2) b;
    cx a,b;
    u1(lambda/2) b;
}
```

pyzx's `ZPhase` gate and `CRZ` gates are equivalent to, respectively, OpenQASM 2's `rz` (a synonym for `u1`) and `crz` gates. Confusingly, the `crz` gate is not simply a controlled version of the `rz` gate; that gate is named `cu1`.

Meanwhile, the OpenQASM 3 spec defines:
```
gate rz(λ) a { gphase(-λ/2); U(0, 0, λ) a; }
gate crz(θ) a, b { ctrl @ rz(θ) a, b; }
```

This has the advantage that `crz` is now simply a controlled version of `rz`, at the expense of changing the meaning of `rz` so that it no longer matches `u1`/`ZPhase`.

This difference in conventions leads to user confusion and coding errors, as seen in issues #102 and #103. As well, changing the behaviour of `qasmparser` depending on the version of OpenQASM used is a blocker for issue #116.

### DIFF
--- a/tests/test_circuit.py
+++ b/tests/test_circuit.py
@@ -31,6 +31,13 @@ try:
 except ImportError:
     np = None
 
+try:
+    from qiskit.circuit import QuantumCircuit
+    from qiskit import quantum_info
+    import math
+except ImportError:
+    QuantumCircuit = None
+
 from pyzx.generate import cliffordT, cliffords
 from pyzx.simplify import clifford_simp
 from pyzx.extract import extract_circuit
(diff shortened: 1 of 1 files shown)


dlyongemallo:
After investigating, it turns out OpenQASM's `rz` is supposed to have a global phase (and be identical to how `rz` for OpenQASM 3) in principle, but there just wasn't a way to express that in code in OpenQASM 2. This means that the existing behaviour of pyzx when it comes to importing/exporting OpenQASM 2 files is technically incorrect, though when a file represents a complete circuit a global phase doesn't matter. Continuing to investigate whether (and how) the behaviour of pyzx needs to be changed.

dlyongemallo:
This PR is superseded by #156.

Issue #87, opened 2022-04-26
input qasm file format


anitalu724:
Hi, thanks for the useful tool for implementing ZX-calculus.
I'm wondering if it is possible to enter all formats of qasm files, since I found out that some gates like P-gate, CP-gate, and so on are not supported by the `load` function provided in this tool. Thanks~

jvdwetering:
That would certainly be possible. Is there a specific library that uses these gates? I don't really know what a 'P' gate is namely (I'm assuming 'CP' is just a controlled-P gate)

anitalu724:
In my viewpoint, qasm format is kind of free style. However,  in the tool Qiskit provided by IBM, there are a couple of gate definitions. [Qiskit Gate Library](https://qiskit.org/documentation/apidoc/circuit_library.html) The output qasm files of Qiskit contain some of these gates which are not supported in PyZX.

jvdwetering:
That list you link does not mention a P gate, unless I missed it?

anitalu724:
Sorry, I didn't explain it clearly enough. P-gate is equivalent to [PhaseGate](https://qiskit.org/documentation/stubs/qiskit.circuit.library.PhaseGate.html#qiskit.circuit.library.PhaseGate) in the [Qiskit Gate Library](https://qiskit.org/documentation/apidoc/circuit_library.html). Some papers and tutorial called PhaseGate "P-gate" for short.

dlyongemallo:
Hi, @anitalu724,

I am working on #156 which is going to add support for most of the OpenQASM standard library gates. Besides `p` and `cp`, was there anything else you needed?

dlyongemallo:
As of PR #156 both `p` and `cp` are supported, along with the most common qasm gates, so unless there are other requested gates, this issue can be closed.

PR #156, opened 2023-09-16
Add support for most OpenQASM 2 and 3 standard gates.

### LINKED ISSUE #153: cp gate support for QASMParser
Hi @jvdwetering, I faced the problem of phase(p), and controlled phase(cp) gate when parsing the qasm string. I thought I might try working on this issue to add p and cp gate support. Your guidance and input would be greatly appreciated.

dlyongemallo:
Adds support for the most common OpenQASM 2 and 3 standard library gates, and a unit test to ensure that the implementations in pyzx and qiskit are the same up to a global phase.

This PR fixes #153 and #158, adds tests to verify #102 and #103, and should now go most of the way towards completing #87 and #116.

### DIFF
--- a/pyzx/circuit/gates.py
+++ b/pyzx/circuit/gates.py
@@ -498,6 +498,22 @@ class Y(YPhase):
     def __init__(self, target: int) -> None:
         super().__init__(target, phase = Fraction(1,1))
 
+class CY(Gate):
+    name = 'CY'
+    qasm_name = 'cy'
+    def __init__(self, control: int, target: int) -> None:
+        self.target = target
+        self.control = control
+
+    def to_basic_gates(self):
+        return [S(self.target,adjoint=True),
+                CNOT(self.control,self.target),
+                S(self.target)]
+
+    def to_graph(self, g, q_mapper, c_mapper):
+        for gate in self.to_basic_gates():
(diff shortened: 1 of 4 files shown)


dlyongemallo:
@jvdwetering I'm slightly unsure about whether `to_basic_gates()` for `crx` and `rxx` are correct. Does the list which is returned have to consist only of gates for which `to_basic_gates()` return themselves, or is `to_basic_gates()` defined recursively?

dlyongemallo:
There are a few outstanding gates in `qelib1.inc` and `stdgates.inc` which haven't been included. I wanted to make sure that these are correct first, before I implement those.

jvdwetering:
to_basic_gates should itself return a list of basic gates, so it should not have to be called recursively.

jvdwetering:
I've never seen anyone use those gates, so I think it's fine if these are not implemented for now.

jvdwetering:
Can this be merged?

dlyongemallo:
AFAIK, yes. Unless you have any comments or requests about the code.

dlyongemallo:
p.s. I just want to thank @jakelishman for his guidance in explaining the difference between how OpenQASM 2 and 3 represents the gates, which made this PR possible.

PR #164, opened 2023-10-01
Z box

RazinShaikh:
Basic support for the Z box generator. You can create, rewrite, draw, tensor contract, and save graphs with Z boxes. I have added the fusion rewrite, which works between two Z boxes or a Z box & Z spider. I have also added rewrite to convert a Z spider to Z box. Additional rewrites can be added in the future but that is conceptually straightforward.


jvdwetering:
Very nice! Is this ready to be merged?

RazinShaikh:
I think it is but if you think I have missed something, please let me know.

jvdwetering:
Maybe at some point we want to have a seperate file for all the ZXW rewrites, just like how all the H-box rules are in `hrules.py` and `hsimplify.py`, but that probably only makes sense once there are more rewrite rules.

RazinShaikh:
Yeah that would be good. For some rewrites, the Z box behaves almost the same as the Z spider with a slight modification on how phases are modified. For those, I would prefer to keep it in the same rewrite.

PR #173, opened 2023-10-21
Make `RZZ` a subclass of `ParityPhase`.

dlyongemallo:


dlyongemallo:
I'm not sure this is strictly necessary. The only thing it adds is it makes comparison between `RZZ` and `ParityPhase` possible, but I doubt that comes up often. And maybe conceptually it makes the relationship between them clearer.

PR #175, opened 2023-10-30
Make the docstring for `graph_to_swaps` a bit clearer.

dlyongemallo:


dlyongemallo:
See https://github.com/Quantomatic/pyzx/issues/161#issuecomment-1742023960 for context.

PR #188, opened 2023-11-28
Catch up CHANGELOG.

dlyongemallo:
I noticed that the CHANGELOG was quite behind, so I caught it up to today.  Less work to do later for whoever releases the next version.

Based on previous entries, it seems that changes made by @jvdwetering and @akissinger are not specifically credited by name. Not sure if anyone else is included in that group, so I credited everyone else by name. Let me know if I missed anyone/anything or if anything needs to be changed.

Also, I noticed that @ABorgna has been credited previous under both that name and @aborgna-q (seems to be personal vs work account). The specific change noted in this changelog update was under the @ABorgna handle, so I've used that, but let me know if it should be under the other one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,34 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Right now this project is in Beta and does not yet follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html). 
 Hence, occasionally changes will be backwards incompatible (although they will all be documented here).
 
+## [0.x.x] - xxxx-xx-xx
+
+
+### Added
+- Support for W nodes and Z boxes (courtesy of @RazinShaikh).
+- Support for poly phases (courtesy of @RazinShaikh)
+- Support for Jupyter notebooks in documentation using nbsphinx (courtesy of @dlyongemallo).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,34 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Functions `simplify.to_clifford_normal_form_graph` and `extract.extract_clifford_normal_form`.
+- Lazy import of some dependencies.
+
+### Changed
+- Class `CX`, which refers to an X-controlled X gate, renamed to `XCX` for clarity.
+- Parameters for `FSim` changed to put control and target before angles, for consistency with other gates.
+
+### Fixed
+- A bunch of mypy issues.
+- Variable types now saved in json.
```
jvdwetering: Change to 'json export and import supports Poly phases', because this feature didn't exist at all before.

jvdwetering:
Thanks! I was meaning to do this, and then other stuff just kept getting in the way. 
I think we should also add that there has been a breaking change in how the json gets saved, which means older json is no longer parsible (this is so that now the json correctly remembers input/output ordering).

dlyongemallo:
Made the requested changes.

PR #205, opened 2024-03-30
Call replacement functions from deprecated functions.

### LINKED ISSUE #203: `Graph.to_json` followed by `pyzx.json_to_graph` produces an error
This example:
```python
graph = zx.Graph()
graph.add_vertex(zx.VertexType.Z, phase=1)
zx.json_to_graph(graph.to_json())
```
throws an error:
```
File (...)/python3.11/site-packages/pyzx/io.py:84, in json_to_graph(js)
     [82] else: raise TypeError("unsupported type '{}'".format(d['type']))
     [83] if 'value' in d:
---> [84]     g.set_phase(v,_quanto_value_to_phase(d['value']))
     [85] else:
     [86]     g.set_phase(v,Fraction(0,1))

File [(...)/python3.11/site-packages/pyzx/io.py:39], in _quanto_value_to_phase(s)
     [37]     except ValueError:
     [38]         raise ValueError("Invalid phase '{}'".format(s))
---> [39] return Fraction(s)

File (...)/python3.11/fractions.py:115, in Fraction.__new__(cls, numerator, denominator, _normalize)
    [113] m = _RATIONAL_FORMAT.match(numerator)
    [114] if m is None:
--> [115]     raise ValueError('Invalid literal for Fraction: %r' %
    [116]                      numerator)
    [117] numerator = int(m.group('num') or '0')
    [118] denom = m.group('denom')

ValueError: Invalid literal for Fraction: 'π'
```

Weirdly, any other combination of `Graph.to_json`/`graph_to_json` and `Graph.from_json`/`json_to_graph` works correctly.

This is the json from `Graph.to_json()`:
```json
{"wire_vertices": {}, "node_vertices": {"v0": {"annotation": {"coord": [-1, 1]}, "data": {"type": "Z", "value": "\u03c0"}}}, "undir_edges": {}, "variable_types": {}, "scalar": "{\"power2\": 0, \"phase\": \"0\"}"}
```

And this is the one from `json_to_graph`:
```json
{"wire_vertices": {}, "node_vertices": {"v0": {"annotation": {"coord": [-1, 1]}, "data": {"type": "Z", "value": "\\pi"}}}, "undir_edges": {}}
```

The former uses an unicode character to encode π (and also seems to always include the scalar).

- I'm using `pyzx=0.8.0`

dlyongemallo:
This may be a breaking change for users who are calling the deprecated functions. Presumably, though, this change actually produces the correct intended behaviour for these calls.

To force the old behaviour, there is an optional parameter `force_deprecated_behavior` (defaults to `False`) which can be set to `True`.

Fixes #203.

### DIFF
--- a/pyzx/io.py
+++ b/pyzx/io.py
@@ -1,4 +1,4 @@
-# PyZX - Python library for quantum circuit rewriting 
+# PyZX - Python library for quantum circuit rewriting
 #        and optimization using the ZX-calculus
 # Copyright (C) 2018 - Aleks Kissinger and John van de Wetering
 
@@ -21,7 +21,6 @@ from typing import List, Dict, Any
 from .utils import FractionLike
 from .graph import Graph, EdgeType, VertexType
 from .graph.base import BaseGraph, VT, ET
-from .simplify import id_simp
 from .symbolic import Poly
 
 __all__ = ['json_to_graph', 'graph_to_json', 'to_graphml']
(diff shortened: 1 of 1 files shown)


dlyongemallo:
Is there a way to notify users of the library? It would be good to be able to communicate deprecation notices to users.

Issue #198, opened 2024-02-23
Error in full_optimize


vsprasannaa:
Dear All, 

I tried full_optimize for a circuit. Before doing this, when we check for circuit stats this way: 

`print(pyzx_circ.stats())`

we get: 

Circuit EvolvedOps-2842 on 10 qubits with 450 gates.
        30 is the T-count
        420 Cliffords among which
        274 2-qubit gates (274 CNOT, 0 other) and
        0 Hadamard gates.

But upon doing a full_reduce followed by full_optimize this way: 

```
g = pyzx_circ.to_graph()
zx.full_reduce(g, quiet=False) 
g.normalize()
```

```
c_opt = zx.extract_circuit(g.copy(),optimize_czs=True, optimize_cnots=3, up_to_perm=False, quiet=True)
zx.full_optimize(c_opt.to_basic_gates(), quiet=True)
```

I run into this error: 

TypeError: This method only works on Clifford+T circuits. This circuit contains a ZPhase(4,phase=940347/1909922) 

Can someone please help resolve this issue? Thanks in advance! :-)

jvdwetering:
It means your circuit contains phase gates that are not a multiple of pi/4, and hence not supported by full_optimize. You can use basic_optimize instead.

vsprasannaa:
Thank you very much for the clarification!

PR #206, opened 2024-04-06
Add helpful suggestion when calling `full_optimize` with non-Clifford+T circuit.

dlyongemallo:
See #198 for context.


jvdwetering:
I agree with the idea, but the way the wording is phrased might make it seem that `basic_optimization` can make a non-Clifford+T circuit Clifford+T, so that then this function can be called. Maybe the wording can be changed to something like 'for these circuits stick to `basic_optimization`.

dlyongemallo:
Done.

PR #216, opened 2024-04-22
Vertex/edge cutting

mjsutcliffe99:
Included functions for applying the vertex and edge cutting decompositions

### DIFF
--- a/pyzx/simulate.py
+++ b/pyzx/simulate.py
@@ -542,3 +542,35 @@ def cut_edge(g,e,ty=1):
     g1.add_edge((v,e[1]),etype)
     
     return (g0,g1)
+
+def cut_wishbone(g,v,neighs,ph):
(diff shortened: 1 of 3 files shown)


mjsutcliffe99:
> Can you please add tests?

I've just added some tests for both functions to verify the scalar results are correct on randomly generated graphs

jvdwetering:
Could you add docstrings to the main functions you added, with perhaps a reference to your papers where you use them.

jvdwetering:
Ah I see what goes wrong. You are using match statements, but these were introduced in Python 3.10. PyZX still supports older versions of Python. I see that in that match statement you are also using hard-coded `1` and `2` for the types of Z and X spiders. You can actually replace this by calling the function `utils.toggle_vertex` for the same functionality.

jvdwetering:
Can this be merged?

Issue #123, opened 2023-07-30
Multigraph support


akissinger:
I've been thinking about the least intrusive way to add support for parallel edges and self-loops to pyzx, and doing some experimental work on the [multigraph](https://github.com/Quantomatic/pyzx/tree/multigraph) branch.

I haven't thought of all the details yet, but here are some notes (included breaking changes). Feedback welcome.

- multigraph support is optional, and depends on the graph backend. If a backend has multigraph support, it should override the new `multigraph` method of BaseGraph to return True
  - if a method hasn't (yet) been updated to cope with multigraphs, it should throw a NotImplementedError if it gets passed one
  - I've made a copy of GraphS and called it Multigraph. This will be the pure python backend modified to have multigraph support
- BaseGraph should be cleaned up so that its utility methods work for graphs OR multigraphs (in progress)
  - If necessary, methods can check `self.multigraph()` and implement different behaviors if this is true (although this doesn't seem to be necessary for the most part)
  - I think this is a good time to deprecate some overly complicated methods like `replace_subgraph`
  - `add_edge_table` is now just iteratively calling `add_edge_smart`, which should be implemented on the backend to Do The Right Thing, depending on whether the graph is a multigraph or not.
  - `ET` should be treated as an opaque ID, rather than a source/target pair. This was always the intention, but it got lost at some point.
- `Graph.edge(s, t)` now returns the ID of the *first* edge connecting `s` and `t`. Its behavior is undefined if `s` and `t` are not connected.
- `Graph.edges` now has optional `s` and `t` parameters to get a Sequence of all the edges connecting `s` and `t`. If these are omitted, it gives a Sequence of all edges, as before.

dlyongemallo:
I'm not familiar with the design decisions that went into the current implementation of graphs in pyzx, but what's the reason for using adjacency lists over an adjacency matrix? In the case of multigraph support, with an adjacency matrix one could just use a weight for each edge type. For example if you had 3 (simple) edges, you'd just give the edge a value of 3, instead of having to manipulate 3 edges. This scales better if one can have many edges between two vertices. The tradeoff is that one can then no longer manipulate any individual edge, if that's really important for some reason.

jvdwetering:
[...]

PR #217, opened 2024-05-03
Refactoring to support multigraph backend

akissinger:
I needed to do some non-trivial refactoring to get PyZX ready for multigraph support (cf. issue #123).

The multigraph backend is still not completely ready, but I think now might be a good time to commit my refactoring to minimise conflicts going forward, since this touches almost every file. The main issue is to revert `ET` to being an abstract type representing the identity of an edge, rather than an alias to `Tuple[VT,VT]`, because the latter no longer refers to a unique edge in the graph.

Major changes are:

* `ET` is now an opaque type, not an alias for `Tuple[VT,VT]`. If I have an `e` of type `ET`, I should use the relevant graph methods to extract it's source and target, rather than simply calling `e[0]` and `e[1]`.
* `add_edge` and `add_edge_list` should take arguments of type `Tuple[VT,VT]`, not `ET`
* `add_edge` in GraphS is now always "smart", i.e. it will try to handle parallel edges or throw an error. The old "dumb" version, where parallel edges get silently overwritten is gone.
* `add_edge_table` just repeated calls `add_edge`, so all the parallel edge logic has been shifted to `add_edge`
* there is a new function `upair` defined in `BaseGraph` that returns a given pair of vertices as an "unordered pair", i.e. in a canonical order. This is what `GraphS.edge` does, but we should no longer assume *all* backends will do this.

Note that some temporary hacks to get H and W nodes to work with parallel edges have not survived the refactor. I think it would be better instead to get rid of these hacks and assume with a working with multigraphs whenever we need parallel edges and non-ZX nodes. Unfortunately, this might temporarily break some ZXlive stuff.


RazinShaikh:
Exciting to see that multigraph is finally done. I can help in integrating it with ZXLive after the PR is merged.

dlyongemallo:
Is there any way to break this down into smaller, logically independent pieces, or is everything tied together in such a way that it would be impossible to separate?

Can you add tests, especially where the multigraph backend behaves differently than the existing simple-graph one (ideally testing both backends to show the difference)? This is especially useful for catching edge cases.

akissinger:
re: breaking in to chunks: not really. I just wanted to flag this up with a few other people to do a rough sanity check before I merge. This is a major change, so if some stuff breaks, we can deal with it as and when we find it.

re: multigraph tests, those should be written eventually, but I think it's a good idea to go ahead and merge sooner rather than later because this touches so many files. Maybe it's a good idea for the multigraph backend to issue some kind of warning to use at your own risk until it has been tested a bit more extensively.

pyzx/graph/diff.py
```diff
@@ -80,13 +84,13 @@ def calculate_diff(self, g1: BaseGraph[VT,ET], g2: BaseGraph[VT,ET]) -> None:
...
 				self.changed_pos[v] = pos2
 
-		for e in new_edges:
-			if e in old_edges:
-				if g1.edge_type(e) != g2.edge_type(e):
-					self.changed_edge_types[e] = g2.edge_type(e)
-			else:
-				if g2.edge_type(e) != EdgeType.HADAMARD: # We take Hadamard edges to be the default
-					self.changed_edge_types[e] = g2.edge_type(e)
+		# for e in new_edges:
```
dlyongemallo, line 87: Should this just be deleted? I noticed that there are a couple of places in pyzx where blocks of code are commented out without explanation. If these should be kept, it would be helpful to readers to have a preceding comment along the lines of "Alternate implementation of ..." or "Uncomment this block if you want to...". Otherwise, for someone new to a codebase who encounters blocks of code which are commented out, it's just puzzling why they're there.

pyzx/graph/multigraph.py
```diff
@@ -0,0 +1,380 @@
...
+                    if v1 > v0:
+                        for _ in range(e.s): yield (v0, v1, EdgeType.SIMPLE)
+                        for _ in range(e.h): yield (v0, v1, EdgeType.HADAMARD)
+        elif t != None:
+            s, t = (s, t) if s < t else (t, s)
+            e = self.graph[s][t]
+            for _ in range(e.s): yield (s, t, EdgeType.SIMPLE)
+            for _ in range(e.h): yield (s, t, EdgeType.HADAMARD)
+
+    # def edges_in_range(self, start, end, safe=False):
```
dlyongemallo, line 266: Ditto.

pyzx/graph/multigraph.py
```diff
@@ -0,0 +1,380 @@
...
+    #            if all(start<v1<end for v1 in adj) and all(all(start<v2<end for v2 in self.graph[v1]) for v1 in adj):
+    #                for v1 in adj:
+    #                    if v1 > v0:
+    #                        yield (v0,v1)
+
+    def edge(self, s, t):
+        return (s,t) if s < t else (t,s)
+    def edge_set(self):
+        return set(self.edges())
+    def edge_st(self, edge):
```
dlyongemallo, line 292: `edge_set` and `edge_st` are dangerously close to "minor typo leads to frustrating-to-find bug" territory, though fortunately they have different signatures.

dlyongemallo:
> I think it's a good idea to go ahead and merge sooner rather than later because this touches so many files.

I don't see any major issues with the changes, and rather than block this on a more in-depth review, I don't object to merging this now. (I have some minor issues with places where more type hints would be helpful, or where blocks of code are commented out without explanation, but the existing codebase is already like this, and these can be fixed en masse later.)

One thing I would strongly urge though is to change the commit message to be something like "[BREAKING CHANGE] Refactor to support multigraph backend" (per my comment https://github.com/Quantomatic/zxlive/issues/229#issuecomment-1913075173).

akissinger:
Cool, done. I added [BREAKING CHANGES] to the commit msg for the merge commit.

PR #223, opened 2024-05-05
Fix `pyzx.draw()` causing an error due to missing subplotspec in some matplotlib environments.

### LINKED ISSUE #222: `pyzx.draw()` does not work in some matplotlib environments
I am embedding a ZX diagram into a web app (using [shiny](https://shiny.posit.co/py/)), and calling `pyzx.draw(g)` results in an error `'NoneType' object has no attribute 'rowspan'`. 

The cause is due to the way that PyZX is creating the figure in [these lines](https://github.com/Quantomatic/pyzx/blob/master/pyzx/drawing.py#L178-L179), which does not set [`matplotlib.gridspec.SubplotSpec`](https://matplotlib.org/stable/api/_as_gen/matplotlib.gridspec.SubplotSpec.html) properly. 

```python
    fig1 = plt.figure(figsize=figsize)
    ax = fig1.add_axes([0, 0, 1, 1], frameon=False)
```

Some matplotlib environments ignore this silently, but in others this causes an error. For example, this fails also in the latest version of pandas (see [this issue](https://github.com/pandas-dev/pandas/issues/54577) and [this Stack Overflow answer](https://stackoverflow.com/questions/75824264/issues-when-plotting-with-axes-attributeerror-nonetype-object-has-no-attribut/76913512#76913512)).

This can be fixed by calling `subplots` instead:

```python
    fig1, ax = plt.subplots(figsize=figsize)
    ax.set_frame_on(False)
```

But I think this might result in a slightly different appearance in some environments, as the axes are computed rather than explicitly set. Were the original values `[0, 0, 1, 1]` explicitly chosen to force a particular appearance? Otherwise, it seems it's likely more robust to let them be computed based on the environment.

dlyongemallo:
Fixes #222.

### DIFF
--- a/pyzx/drawing.py
+++ b/pyzx/drawing.py
@@ -175,8 +175,9 @@ def draw_matplotlib(
 
     if isinstance(g, Circuit):
         g = g.to_graph(zh=True)
-    fig1 = plt.figure(figsize=figsize)
-    ax = fig1.add_axes([0, 0, 1, 1], frameon=False)
+    fig1, ax = plt.subplots(figsize=figsize)
+    ax.set_frame_on(False)
+    ax.set(xlim=(0, 1), ylim=(0, 1))
     ax.xaxis.set_visible(False)
     ax.yaxis.set_visible(False)
     vs_on_row: Dict[FloatInt, int] = {} # count the vertices on each row


dlyongemallo:
**Code review note:** This PR should be tested by a variety of users in different environments to ensure the appearance of the diagrams are the same as before. I've manually tested it in different environments on my own machine, and in theory I've set everything which had previously been explicitly set to their existing values, but the call to `subplots` is now setting the `matplotlib.gridspec.SubplotSpec` which was previously not set before and could therefore potentially change the figures' appearance in some environments.

dlyongemallo:
Has anyone had a chance to look at this? I'm pretty sure this fixes the issue in those environments for which missing subplotspec causes a problem, and is a no-op otherwise.

RazinShaikh:
I tried this and the figures look same to me.

dlyongemallo:
Can this be merged please?

Issue #229, opened 2024-05-30
Clifford simp adds a phase of n*pi/2


rafaelha:
In the following code snippet, the two tensors differ by a phase of 1j. Running this for different seeds, I find that the phase of `z` is either [1, 1j, -1, -1j].

```
import numpy as np
import random
import pyzx as zx

random.seed(1)
g1 = zx.generate.cliffordT(3, 20, p_t=0)
g2 = g1.copy()

zx.clifford_simp(g2)  # Removing this line makes the assert pass
assert np.allclose(g1.to_tensor(), g2.to_tensor())

z = g1 + g2.adjoint()
zx.clifford_simp(z)
# print(z.to_tensor())
assert z.is_id()
assert z.scalar.to_number() == 1, f"scalar is {z.scalar.to_number()}" 
# >> AssertionError: scalar is (-1.8369701987210297e-16-1j)
```

version: pyzx 0.8.0

rafaelha:
Looking into this further, the problem was simply that `BaseGraph.adjoint()` did not conjugate the phase. Adding the line 
`g.scalar.add_phase(- 2 * g.scalar.phase)` in `BaseGraph.copy` fixed everything.

jvdwetering:
I think this can now be closed?

PR #230, opened 2024-05-30
Fix bug where scalar phase was not conjugated by BaseGraph.adjoint()

rafaelha:
Fixes issue #229

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -30,6 +30,7 @@ from pyzx.generate import identity
 
 import numpy as np
 from pyzx.tensor import compare_tensors
+from pyzx.graph.scalar import Scalar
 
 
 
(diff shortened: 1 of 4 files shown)


jvdwetering:
I'm afraid the fix will be a bit more complicated than that. A Scalar has a `phase`, a list of `phasenodes` and a `floatfactor`. To calculate the conjugate of a Scalar, all of these must be conjugated. This is probably best done by a new method in Scalar, `conjugate()` that does that for you.

tests/test_graph.py
```diff
@@ -176,6 +188,7 @@ def setUp(self):
         g.set_outputs((o1, o2))
         g.add_edges([(i1,v),(i2,w),(v,w),(v,o1),(w,o2)])
         self.i1, self.i2, self.v, self.w, self.o1, self.o2 = i1, i2, v, w, o1, o2
+        print(self.graph.scalar.to_number())
```
dlyongemallo, line 191: Was this a debugging statement which was left in accidentally?

It causes the output of the test to look like this:
```
..................s............(1+0j)
.(1+0j)
.(1+0j)
.(1+0j)
.........................sssss.............................
----------------------------------------------------------------------
Ran 93 tests in 3.109s

OK (skipped=6)
```

rafaelha: Sorry, yes I left it accidentally. I saw that you already removed it, thanks!

Issue #122, opened 2023-07-30
enforce PEP8 style with a tool such as pylint or autopep8


dlyongemallo:
Is there any reason not to enforce PEP8 (or any style, as long as it's consistent) on the code? It would make using certain development tasks, such as searching with something using regular expressions, somewhat easier.

dlyongemallo:
(Would anyone seriously object if I started sending PRs to better align the code with PEP8 recommendations?)

jvdwetering:
I'm a little against touching code that doesn't need to be touched, but not very strongly. If you want to put in the time to make it follow the convention, then go ahead. Although I do think that the restriction to 80 character long lines is too strict. On the other hand, if a line is that long that might be a sign that the code needs to be refactored to just not be that long.

dlyongemallo:
Just out of curiosity, what development environment (OS, IDE, etc.) do the main contributors to this codebase (and, incidentally, zxlive) use? I'm mainly using PyCharm in Linux, but I also have access to Windows and VS Code. With modern IDEs there are a number of nice features like detecting common errors and suggested refactorings which implicitly depend on PEP8 style (e.g., because they use regular expression matching). 

A unified style also makes code easier to read, e.g., `if not b in list_of_bools` can be easily misunderstood (the recommended style is `if b not in list_of_bools`).

jvdwetering:
I don't know about the other contributors, but I wrote most of this code in Sublime, although I have now semi-switched to Visual Studio.

PR #233, opened 2024-06-10
Setup formatter

### LINKED ISSUE #232: Coding style
I'm now planning to contribute to this repo., but before I proceed, I would like to ask about the coding style.

- Is there any kind of style guideline for this repo?
- If any, is it different from the Python standard (PEP8)?
- Do you have any plans for using formatter like `black` or `ruff format` ?

EarlMilktea:
Resolves #232 .

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,38 @@
+name: format
+
+on:
+  push:
+    branches: [master]
+  workflow_dispatch:
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.10
+
+      - name: Install ruff
+        run: |
+          python -m pip install --upgrade pip
+          pip install ruff
+
+      - name: Run ruff format
+        run: ruff format .
+
+      - name: Run ruff isort
+        run: ruff check --select I --fix .
+
+      - name: Commit changes
+        run: |
+          date > generated.txt
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m ":art: Format" || exit 0
+          git push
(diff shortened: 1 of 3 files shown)


EarlMilktea:
MEMO: This config. is optional (defaults to `88`)

```toml
line-length = 120
```

dlyongemallo:
> Resolves #232 .

Resolves #122.

akissinger:
I'm not sure about this, mainly just because I've never used auto-formatting at this level before.

It looks like this could potentially spam the commit logs by creating a double commit for almost everything. One is the actual commit, then a second commit e.g. putting a bit less/more whitespace somewhere because someone's editor isn't configured in the same way as ruff. Is consistent formatting worth the extra commit spam and/or effort? (genuine question, not a rhetorical one)

EarlMilktea:
@akissinger (CC: @jvdwetering )

Let me first note that formatter CI does **NOT** run each time you commit: only once per merged PR because it's tracking master branch only.

Actually I admit Ruff (or black similarly) is sometimes too strict, but at the same time, it helps me a lot because...

- Contributors don't have to care about the coding style or consult coding style document
- Reviewer-friendly as diffs. are minimized, also easy to grep
- Reviewers don't have to manually check that the codes are PEP8-compliant
- Can detect unconventional styles: for example

```python
if x: break
```

- Simplifies linter/editor configuration since we don't have to manually inspect which rules are associated with PEP8 

If you're not happy with automatic commit, I'm willing to rewrite the CI to just report problems and urge contributors to fix.
Which do you prefer?

dlyongemallo:
> Is consistent formatting worth the extra commit spam and/or effort? (genuine question, not a rhetorical one)

In a large code base touched by many people, a consistent style make certain tasks easier or more pleasant. Reading unfamiliar code in which the style switches between code blocks imposes an unnecessary cognitive load because the reader has to think about whether apparent differences are meaningful.

Extra commits can probably be reduced by documenting the expected style in the contributor guidelines and encouraging contributors to use automatic formatting tools (e.g., format on save in their editor).

dlyongemallo:
@EarlMilktea Why was this closed? I'd like to see something like this implemented.

PR #234, opened 2024-06-12
fixing issue #234 of ZXLive

lia-approves:
For Poly, there is not a sensible value to set denominator to, so this checks if the phase type is Poly before checking the denominator value. Addresses https://github.com/Quantomatic/zxlive/issues/234
With @hm919

### DIFF
--- a/pyzx/graph/jsonparser.py
+++ b/pyzx/graph/jsonparser.py
@@ -46,6 +46,7 @@ def string_to_phase(string: str, g: Union[BaseGraph,'GraphDiff']) -> Union[Fract
         return Fraction(0)
     try:
         s = string.lower().replace(' ', '')
+        s = s.replace('*', '')
         s = re.sub(r'\\?(pi|\u03c0)', '', s)
         if s == '': return Fraction(1)
         if s == '-': return Fraction(-1)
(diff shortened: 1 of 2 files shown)


pyzx/graph/jsonparser.py
```diff
@@ -54,7 +55,7 @@ def string_to_phase(string: str, g: Union[BaseGraph,'GraphDiff']) -> Union[Fract
         elif '/' in s:
             a, b = s.split("/", 2)
             if not a:
-                return Fraction(1, int(b))
+                return Fraction(int(1), int(b))
```
dlyongemallo, line 58: Is this just for visual consistency (to aid readers of the code), or does this change actually have an effect?

lia-approves: This change from 1 to int(1) _actually has an effect_. For some reason I don't understand, I presume some obscure typing convention that could be platform dependent, there is a bug that setting the phase of a node in ZXLive to "2a+b" automatically converts to "0a+b". Oddly, changing this line here fixes this so that it actually stays as "2a+b". This was the case on at least two devices. I don't know where if there is some typechecking somewhere in the codebase or elsewhere to see if something is of type int, or why python doesn't by default have 1 be of type int, or maybe there's different int types,...

dlyongemallo: Oh, wow. This warrants a comment in the code with this explanation so that it isn't accidentally removed by a person or tool later which mistakenly thinks this is a noop.

PR #236, opened 2024-06-12
several fixes and improvements to multigraph

RazinShaikh:
These fixes were found while adding multigraph support in ZXLive. It would be nice to merge this sooner as the ZXLive main branch is broken right now.


RazinShaikh:
That line is a bit hacky. I can change it to Counter.

RazinShaikh:
I didn't test properly when I changed to Counter. I have fixed a bug where the correct way of looping through Counter is to call Counter.elements(). It is fixed in PR https://github.com/zxcalc/pyzx/pull/241.

PR #239, opened 2024-06-22
Fix multigraph color change

boldar99:
The `color_change` function had been broken with the `Multigraph` implementation.

Changes:

- The implementation of `color_change` has been modified to call a new function that toggles between Hadamard and Simple edge types.

- The `toggle_edge` utility function has been changed not to do anything when the input edge type is neither Hadamard nor Simple.

- Implement `__repr__` for `multigraph.Edge` for better error messages

Related to zxcalc/zxlive#256

### DIFF
--- a/pyzx/editor_actions.py
+++ b/pyzx/editor_actions.py
@@ -53,8 +53,7 @@ def color_change(g: BaseGraph[VT,ET], matches: List[VT]) -> rules.RewriteOutputT
     for v in matches:
         g.set_type(v, toggle_vertex(g.type(v)))
         for e in g.incident_edges(v):
-            et = g.edge_type(e)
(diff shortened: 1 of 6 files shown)


jvdwetering:
I'm not sure if we should maybe play it safe, and raise an exception if `toggle_edge` is called on an edge that we don't expect to be toggleable, instead of just silently doing nothing.

boldar99:
That's a good point! I just pushed a fix.

RazinShaikh:
I am not sure if toggling the edge type for multigraph should simply flip the number of simple and Hadamard edges. It makes sense for colour change but what if in a different context, switching the edge type of only the chosen edge (not all edges) makes sense?

PR #240, opened 2024-06-22
Turn VertexType and EdgeType into enums

boldar99:
VertexType and EdgeType were implemented as data classes with hard-coded types. 
A better solution for implementing these classes is using [Enums](https://docs.python.org/3/library/enum.html). 

Changes:
- Use IntEnums to implement VertexType and EdgeType.
- Change the uses of `VertexType.Type` to `VertexType`. Similarly for `EdgeType`.
- Fix the uses of hard-coded integers where the enums can be used.
- Rename some variables that mypy picked on after the update.


RazinShaikh (approved):
LGTM

jvdwetering:
Are you sure you could all the cases where this has to be updated? Did you do some kind of automated search?

boldar99:
I used PyCharm's automated search and replace to deal with `EdgeType.Type` and `VertexType.Type`.
Now I double-checked it with `grep` as well which had no matches, so it seems good.

PR #244, opened 2024-06-27
Fix color_change, fusion, and match_ids

boldar99:
Fix the `color_change`, `fusion`, and `match_ids_parallel` methods that were broken with incident multi-edges.

Changes:

- `incident_edges` now iterates through all individual incident edges.
- `match_ids_parallel` returns False if the degree of the given vertex is not 2. This stops the program from calling `edge_type` which is undefined for multi-edges.

Related to zxcalc/zxlive#256


RazinShaikh:
This only fixes colour change method in editor_actions.py but not the one in basicrules.py. I don't know why there are two implementations of colour change and why they are implemented differently. I tried to copy the implementation of editor_actions.py to basicrules.py and it seems to work.

RazinShaikh:
I have pushed the color_change fix for basicrules.py.

RazinShaikh:
I have now fixed spider fusion in rules.py. I don't think fusion or color_change is defined anywhere else. So this PR should be good to merge.

RazinShaikh:
ah self loops are still broken

RazinShaikh:
Fixed!

jvdwetering:
After the merge conflict is solved, this can probably be merged.

RazinShaikh:
I have resolved the conflict but I need to check if that broke anything

RazinShaikh:
This can be merged now!

PR #251, opened 2024-07-02
Fixing remove id when removing id results in a self-loop

RazinShaikh:


RazinShaikh:
Remove id works now with multigraph.

PR #254, opened 2024-07-10
return an empty generator when there are no edges between source and target

RazinShaikh:
This fixes some keyerror exceptions

### DIFF
--- a/pyzx/graph/multigraph.py
+++ b/pyzx/graph/multigraph.py
@@ -277,6 +277,9 @@ class Multigraph(BaseGraph[int,Tuple[int,int,EdgeType]]):
                         for _ in range(e.w_io): yield (v0, v1, EdgeType.W_IO)
         elif t != None:
             s, t = (s, t) if s < t else (t, s)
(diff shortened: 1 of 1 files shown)


jvdwetering:
You write
```return
yield```
How does that work? Shouldn't it just be `yield` if you want to return a generator. Or am I misunderstanding how empty generators work. If so, maybe add a comment that this is intentionally returning an empty generator?

RazinShaikh:
I don't really understand why exactly this returns an empty generator. I got it from this stackoverflow answer with 200 votes. Without the yield, the return statement returns None instead of a generator.

https://stackoverflow.com/questions/13243766/how-to-define-an-empty-generator-function

RazinShaikh:
I don't really understand why exactly this returns an empty generator. I got it from this stackoverflow answer with 200 votes.

https://stackoverflow.com/questions/13243766/how-to-define-an-empty-generator-function

RazinShaikh:
I misunderstood the stackoverflow answer. The yield statement is unnecessary; pushed the correction.

PR #261, opened 2024-07-20
Bialgebra in the opposite direction

RazinShaikh:


pyzx/editor_actions.py
```diff
@@ -299,6 +300,86 @@ def bialgebra(g: BaseGraph[VT,ET],
...
+
+    etab: dict = defaultdict(lambda: [0, 0])
+    if edge_type == EdgeType.SIMPLE:
+        etab[upair(new_vertex1, new_vertex2)] = [1, 0]
+    else:
+        etab[upair(new_vertex1, new_vertex2)] = [0, 1]
+    update_etab(etab, new_vertex1, neighbors1, loops1)
+    update_etab(etab, new_vertex2, neighbors2, loops2)
+
+    g.scalar.add_power(-(len(neighbors1)-1)*(len(neighbors2)-1)) #TODO: not sure if this is correct
```
RazinShaikh: I am not completely sure about the global scalar, so I would appreciate if someone else can confirm

jvdwetering: This looks about right to me. It should be the inverse of the scalar in the forward bialgebra rule.

RazinShaikh: Okay, do you know what happens with the self-loops? I am not sure if the self-loop scalar is correct even in the forward case.

RazinShaikh:
The error in the checks seems unrelated to the PR; see https://stackoverflow.com/questions/40183108. Can you run the checks again?

RazinShaikh:
I reran the checks by closing and reopening the PR, and now it passed

pyzx/editor_actions.py
jvdwetering: I've never really thought about applying a strong complementarity when there is a self-loop, so I don't know what should happen with the scalar then.

RazinShaikh: I think this is correct so I have removed the todo comment.

RazinShaikh:
this can be merged if it looks good to you @jvdwetering

PR #271, opened 2024-09-04
Implement cats when Clifford has phase pi

alexkoziell:
- Implement cat decomposition when Clifford spider in state being decomposed has phase $\pi$.
- Fix bug in sign of term produced in `gen_catlike_term`.


dlyongemallo:
Can you please add a test to `test_simulate.py` that fails without your change and passes with it?

alexkoziell:
@dlyongemallo done.

PR #273, opened 2024-09-12
Fix invalid escape sequences

EarlMilktea:
This PR fixes invalid escape sequences, such as `\c` in `\chi`.


EarlMilktea:
@jvdwetering Could you review?

PR #277, opened 2024-11-14
Automatic Graph Drawing for draw_d3

mehatamm:
I added a toggle for the `draw_d3` function in drawing.py for graphs without qubit/row data provided, so instead of being drawn stacked they are drawn after the application of a force-based algorithm that produces reasonably symmetric and visually pleasing results. This uses the helper function `auto_draw_vertex_locs`. The algorithm used is the first found in this textbook chapter:
https://cs.brown.edu/people/rtamassi/gdhandbook/chapters/force-directed.pdf


jvdwetering:
Thanks for the implementation. Can I suggest changing the parameter name to `auto_layout` instead of `auto_draw` which I find a bit more ambiguous. Can you also add to the docstring of that function explaining what this parameter does?

mehatamm:
> Thanks for the implementation. Can I suggest changing the parameter name to `auto_layout` instead of `auto_draw` which I find a bit more ambiguous. Can you also add to the docstring of that function explaining what this parameter does?

I changed the parameter name to `auto_layout` and added a docstring to `draw_d3`. Let me know if there's anything else I should change!

PR #279, opened 2024-12-17
Enhance QASM parser to support multi-character register names

contra-bit:
Prior to this patch, the QASM parser could only parse `creg` or `qreg` declarations with a length of 1 (e.g., `q` or `c`). This made it impossible to define or measure quantum registers with variable names longer than one letter.

For example, the parser would fail to parse a `qreg` named `q1`.

This patch addresses this limitation by enhancing the parser to handle multi-character register names. The key changes are:

1. The `parse_command` method has been updated to handle the `measure` operation separately. It now correctly extracts the target register name and index, as well as the result register name and index, from the `measure` command.

2. The logic for parsing `qreg` and `creg` declarations has been improved to support multi-character register names. The register name and index are now extracted separately, and the parser checks if the register name is valid before processing the declaration.

These changes can be tested by parsing the following QASM code:

```python
zx.qasm("""OPENQASM 2.0;
include "qelib1.inc";
qreg q1[3];
creg c1[1];
creg c2[1];
h q1[1];
h q1[1];
x q1[2];
h q1[2];
id q1[2];
h q1[0];
cx q1[0], q1[2];
h q1[0];
measure q1[0] -> c2[0];
measure q1[1] -> c1[0];
""")
```

The updated parser should now be able to correctly parse the QASM code, including the `qreg` and `measure` operations with multi-character register names.

### DIFF
--- a/pyzx/circuit/qasmparser.py
+++ b/pyzx/circuit/qasmparser.py
@@ -140,7 +140,18 @@ class QASMParser(object):
     def parse_command(self, c: str, registers: Dict[str,Tuple[int,int]]) -> List[Gate]:
         gates: List[Gate] = []
         name, phases, args = self.extract_command_parts(c)
-        if name in ("barrier","creg","measure", "id"): return gates
+        if name in ("barrier","creg", "id"): return gates
+        if name == "measure":
+            target, result_bit = args[0].split(' -> ')
+            # Extract the register name and index separately for both target and result
+            target_reg, target_idx = target.split('[')
+            result_reg, result_idx = result_bit.split('[')
+            # Remove the trailing ']' and convert to int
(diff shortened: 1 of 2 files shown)


contra-bit:
I am unsure why mpy complains. Target qbit and and result_bit are cast into ints. 

These are the arguments from my example for Measurement

```sh
target_qbit 0 is a <class 'int'>
result 0 is a <class 'int'>
target_qbit 1 is a <class 'int'>
result 0 is a <class 'int'>
```

jvdwetering:
You are reusing the variable name 'result_bit' which confuses mypy as you use it as str first and then as int.
And maybe you aren't importing the Measurement class?

PR #286, opened 2025-02-18
Fix `Multigraph.edge` and fix `tikz_to_graph`

boldar99:
`Multigraph.edge` now has the same signature as `BaseGraph.edge` and does what the documentation describes.

This also fixes some simplification routines, e.g. pivot_simp.

### DIFF
--- a/pyzx/graph/multigraph.py
+++ b/pyzx/graph/multigraph.py
@@ -320,8 +320,13 @@ class Multigraph(BaseGraph[int,Tuple[int,int,EdgeType]]):
     #                    if v1 > v0:
     #                        yield (v0,v1)
 
-    def edge(self, s, t):
-        return (s,t) if s < t else (t,s)
+    def edge(self, s, t, et: EdgeType=EdgeType.SIMPLE):
+        s, t = (s, t) if s < t else (t, s)
+        e = self.graph[s][t]
+        if e.get_edge_count(et):
+            return (s, t, et)
+        return next(self.edges(s, t))
+
 
     def edge_set(self):
         return Counter(self.edges())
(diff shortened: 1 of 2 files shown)


jvdwetering:
This looks reasonable. I guess since this just adds an argument, it shouldn't break any of the existing code that was adapted for multigraphs?

boldar99:
I believe so. I did some testing in zxlive and it seemed fine.

jvdwetering:
Thanks. I just pushed to zxlive so that it will read from the latest github version of pyzx instead of the pip version, so that this commit should make a difference for people who update zxlive.

akissinger:
Hmm, I lost track of exactly when this optional argument came in, but its behaviour is a bit weird for multigraphs.

I think the expected behaviour should be: if you drop the argument it should give you the first edge of *any* type. If the type is specified it should give you the first edge of that type, and if there is no edge of that type, throw an exception.

Often how the method gets used is to get an edge from its source and target, then query things about the edge, including its type. This is how it's used e.g. in `BaseGraph.compose`.

Right now, the behaviour if you drop the argument is sortof okay, but almost by accident. It will default to normal edges, but if there are no normal edges, it will give the first edge of any type. However, this behaviour is identical to what happens if you ask for the first normal edge, which is weird.

I think the best solution is to make the edge type argument an `Optional[EdgeType]` with default of `None`. Is there any reason not to change the signature to this everywhere?

(also looping in @RazinShaikh)

RazinShaikh:
The graph.edge for simple graph doesn't do any checks. It simple returns the tuple (s, t). My understanding of this function is that it is like a class method used to create valid format of edges between source and target nodes. It shouldn't care whether there is currently an edge or not.

https://github.com/zxcalc/pyzx/blob/master/pyzx%2Fgraph%2Fgraph_s.py#L261-L262

PR #292, opened 2025-04-12
Switch to `toml` file

### LINKED ISSUE #287: Editable installation error
I just tried to install the editable version of `pyzx` through `pip install -e .` and it gave me the following error. 

```
  DEPRECATION: Legacy editable install of pyzx==0.9.0 from local_file_path/pyzx (setup.py develop) is deprecated. pip 25.1 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

```

I would have created a PR for this, but it's not clear to me which of the provided options is preferable to the maintainers.

purva-thakre:
Fixes #287


jvdwetering:
There's a couple of things missing in this file: 
The manifest files are not included in this toml file
The workflows on github are not updated to use poetry, and instead still use the requirements.txt

There might be other things, but these are the ones I noticed. I could take care of this and make it all work at some point in the future, but it might take a while.

purva-thakre:
> The manifest files are not included in this toml file

TBH I mostly ported over information that was in the deleted `setup.py` file.

> The workflows on github are not updated to use poetry, and instead still use the requirements.txt

Maybe I should have discussed this before adding a PR. Was the goal to get rid of `requirements.txt` too? I was thinking of adding [dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups), but decided against it as you can still use the `txt` files for installation.

PR #302, opened 2025-05-21
graph.edge multigraph bug

RazinShaikh:
we should not call g.edge to create a tuple. I missed the instances in jsonparser that did not call g.add_edge but were still calling g.edge for a similar reason. This should fix it and after this all the tests in ZXLive will pass.


jvdwetering:
I do wonder why all the unit tests in PyZX do seem to pass if this was all going wrong.

RazinShaikh:
Are there any unit tests in PyZX that use multigraph? My guess is that multigraph is never tested so these quirks don't seem to matter.

PR #303, opened 2025-05-27
Implement edge data storage in the graph

RazinShaikh:
Edges can now store data similar to vertices. The immediate motivation for this was to store the curves on the edges in zxlive but this is more applications including scalable ZX, qudit ZX and other variants.

### DIFF
--- a/pyzx/graph/graph_s.py
+++ b/pyzx/graph/graph_s.py
@@ -359,7 +359,7 @@ class GraphS(BaseGraph[int,Tuple[int,int]]):
         self._edata.pop(edge, None)
     def edata_keys(self, edge):
         return self._edata.get(edge, {}).keys()
-    def edata(self, edge, key, default=0):
+    def edata(self, edge, key, default=None):
         if edge in self._edata:
             return self._edata[edge].get(key, default)
         else:
--- a/pyzx/graph/graph_s.py
+++ b/pyzx/graph/graph_s.py
@@ -41,9 +41,9 @@ class GraphS(BaseGraph[int,Tuple[int,int]]):
         self._grounds: Set[int] = set()
 
         self._vdata: Dict[int,Any]                      = dict()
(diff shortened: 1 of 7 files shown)


pyzx/graph/graph_s.py
```diff
@@ -352,3 +354,18 @@ def set_vdata(self, vertex, key, val):
             self._vdata[vertex][key] = val
         else:
             self._vdata[vertex] = {key:val}
+
+    def clear_edata(self, edge):
+        self._edata.pop(edge, None)
+    def edata_keys(self, edge):
+        return self._edata.get(edge, {}).keys()
+    def edata(self, edge, key, default=0):
```
jvdwetering: Would a default of None make more sense?

pyzx/graph/jsonparser.py
```diff
@@ -191,9 +192,15 @@ def graph_to_dict(g: BaseGraph[VT,ET], include_scalar: bool=True) -> Dict[str, A
         d["scalar"] = g.scalar.to_dict()
     d['inputs'] = g.inputs()
     d['outputs'] = g.outputs()
+    # Convert tuple keys to strings for JSON compatibility, replacing EdgeType with its integer value
```
jvdwetering, line 195: I don't understand what this is needed for. Later parsing it by literal_eval is probably also quite slow.
I guess k is an edge object, and so is either a tuple of length 2 or 3.

pyzx/graph/multigraph.py
```diff
@@ -78,6 +78,7 @@ def __init__(self) -> None:
         self._grounds: Set[int]                         = set()
 
         self._vdata: Dict[int,Any]                      = dict()
+        self._edata: Dict[Tuple[int,int,EdgeType], Dict[str, Any]] = dict()
```
jvdwetering, line 81: So it looks like edge data applies here to the full multi-edge, and you can't give values to individual edges between vertices, is that correct?

PR #310, opened 2025-06-04
Allow random seeds

mjsutcliffe99:
Allow (optionally) user-specified seeds for random graph generation. This is helpful (e.g. in benchmarking datasets) as it allows one to regenerate the same graph just by specifying the same seed (provided the dimensions and other parameters are the same). This avoids having to save the graph, e.g. as json, if it will be needed again at a later date.

For instance, the graphs g and h in the following will be equivalent as they share the same seed (and the same dimensions and parameters):

```
q = 10
d = 20
g = zx.generate.cliffordT(q,d,seed=227)
h = zx.generate.cliffordT(q,d,seed=227)
zx.draw(g,scale=20)
zx.draw(h,scale=20)
```

### DIFF
--- a/pyzx/generate.py
+++ b/pyzx/generate.py
@@ -246,10 +246,10 @@ def cnots(qubits: int, depth: int, backend:Optional[str]=None, seed:Optional[int
     g.scalar.add_power(depth)
     return g
 
-def accept(p: float, rand:Optional[random.Random]=random.Random()) -> bool:
+def accept(p: float, rand:random.Random=random.Random()) -> bool:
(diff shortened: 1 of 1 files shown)


jvdwetering:
Could you use random.Random() instead so that the randomness reset is happening internally to the function, and doesn't change the outside randomness?

mjsutcliffe99:
> Could you use random.Random() instead so that the randomness reset is happening internally to the function, and doesn't change the outside randomness?

Yes, no problem. I will change this soon!

PR #322, opened 2025-07-30
properly handle tikz import/export of dummy nodes

RazinShaikh:


RazinShaikh:
That would be ideal but empty nodes are already used by boundary vertices. This is not a problem for export as we can export both dummy and boundary vertices to empty nodes in tikz. But if we want to import tikz, then there is ambiguity in whether to interpret empty as boundary or dummy. Perhaps there is some heuristic we can use as dummy nodes never interact with spiders, and boundary nodes only have one edge and no text.

jvdwetering:
Yes I think you can make some kind of semantic check. If a dummy is connected to a spider, do we want to consider that diagram malformed?

RazinShaikh:
For now we should consider it as malformed. I can see a use case to have (1 input-1 output) dummy nodes acting as an identity to give more flexibility in drawing wires. But that would be a lot of effort for a small feature.

PR #325, opened 2025-08-05
Fix and improve `circuit_to_graph`

boldar99:
* Fix: When two-qubit gates are transformed, a new row is added if the two gates would overlap in any way.

OLD:
image
NEW:
image

* Fix: `InitAncilla` after `PostSelect` now always adds a qubit that's not in use.

OLD:
image
NEW:
image

* Improve: Now `TargetMapper` remembers old labels and qubits are reused accordingly.

* Improve: `InitAncilla` and `PostSelect` now doesn't make a ladder like shape any more.


jvdwetering:
Nice!

PR #344, opened 2025-09-25
Replace types_dict with variable registry to fix symbolic phases

RazinShaikh:
## Problem
  The PyZX library previously used a `types_dict` approach for managing symbolic variable types (boolean vs. continuous), which had several critical issues:

1. **Consistency problems**: Different polynomials containing variables with the same name could have inconsistent types
2. **Deep copy issues**: Variable type information was lost when graphs were deep copied
3. **Reference problems**: Type changes wouldn't propagate to all instances of a variable
4. Symbolic phases broken in **ZXLive**: see issues https://github.com/zxcalc/zxlive/issues/385 , https://github.com/zxcalc/zxlive/issues/332 , https://github.com/zxcalc/zxlive/issues/366

## Solution
This PR replaces the `types_dict` approach with a registry-based pattern that:

1. Associates variables with graphs: Each graph has its own `VarRegistry` to track variable types
2. Maintains name-based identity: Variables with the same name in the same graph always share the same type
3. Preserves types during copying: Deep copied graphs have their variables properly rebound to the new registry
5. Simplifies API: No more need for `freeze()` calls to fix variable types

## Key Changes
* Added `VarRegistry` class to manage variable types centrally per graph
* Modified `Var` class to work with a registry instead of local type storage
* Added `rebind_variables_to_registry()` method to ensure variables are properly bound after copying
* Implemented custom `__deepcopy__` for `BaseGraph` to preserve registry references
* Updated serialization in `GraphDiff` to work with the new registry

## Benefits
* More reliable variable type tracking across complex graph operations
* Correct behavior when deep copying graphs
* Simpler, more intuitive API for working with symbolic variables
* Better consistency between variable instances

### DIFF
--- a/pyzx/graph/base.py
+++ b/pyzx/graph/base.py
@@ -99,6 +99,7 @@ class BaseGraph(Generic[VT, ET], metaclass=DocstringMeta):
         # merge_vdata(v0,v1) is an optional, custom function for merging
         # vdata of v1 into v0 during spider fusion etc.
         self.merge_vdata: Optional[Callable[[VT,VT], None]] = None
+        self.variable_types: Dict[str,bool] = dict() # DEPRICATED - mapping of variable names to their type (bool or continuous)
         self.var_registry: VarRegistry = VarRegistry() # registry for variable types
 
     # MANDATORY OVERRIDES {{{
--- a/pyzx/graph/base.py
+++ b/pyzx/graph/base.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 import math
+import copy
 from fractions import Fraction
 from typing import TYPE_CHECKING, Union, Optional, Generic, TypeVar, Any, Sequence
 from typing import List, Dict, Set, Tuple, Mapping, Iterable, Callable, ClassVar, Literal
@@ -23,6 +24,7 @@ from typing_extensions import Literal, GenericMeta # type: ignore # https://gith
 
 import numpy as np
 
+from ..symbolic import Poly, VarRegistry
 from ..utils import EdgeType, VertexType, toggle_edge, vertex_is_zx
 from ..utils import FloatInt, FractionLike
 from ..tensor import tensorfy, tensor_to_matrix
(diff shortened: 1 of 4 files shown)


pyzx/symbolic.py
```diff
@@ -28,29 +28,39 @@
 from fractions import Fraction
 
 
+class VarRegistry:
+    """Registry to track variable types for a specific graph"""
+    def __init__(self):
+        self.types = {}  # name -> is_bool
```
jvdwetering: Type annotation?

jvdwetering: Like, dict[Var,bool]?

pyzx/symbolic.py
```diff
@@ -67,35 +77,27 @@ def __hash__(self) -> int:
...
+            return False
+        return self.name == other.name
 
     def __copy__(self) -> 'Var':
-        if self._frozen:
-            return Var(self.name, self.is_bool)
-        else:
-            assert isinstance(self._types_dict, dict)
-            return Var(self.name, self._types_dict)
+        return Var(self.name, self.is_bool, self._registry)
```
jvdwetering, line 88: Is it intentional that copying the variable also copies the registry?

pyzx/graph/base.py
```diff
@@ -1031,16 +1039,38 @@ def is_well_formed(self) -> bool:
...
+            phase = self.phase(v)
+            if isinstance(phase, Poly):
+                phase.rebind_variables_to_registry(self.var_registry)
+
+    def __deepcopy__(self, memo):
+        """Custom deepcopy implementation to ensure variable registry is properly handled
+        while using Python's default deepcopy behavior."""
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
```
jvdwetering, line 1060: What is this memo thing? Some deep Python voodoo?

jvdwetering:
Added a couple of comments, but I agree that a solution like this is probably more robust than what we had before.

pyzx/graph/base.py
RazinShaikh: It is indeed deep Python voodoo. From what I understand, memo is a dictionary used in deepcopy which maps id(original_object) → copied_object.

pyzx/symbolic.py
RazinShaikh: Passing the registry object here returns a copy of the variable as new variable object which is part of the same registry. This behaviour makes sense if we want to copy the phase of one spider onto another.

But I am not sure if in general it's better to return a copied variable with the same registry or with a blank new registry. If a graph is copied, we need to rebind the variable to a new registry anyway so this wouldn't make a difference.

pyzx/symbolic.py
RazinShaikh: already added this in the previous commit

PR #348, opened 2025-09-30
Support `reset` and `measure` in `QASMParser`

Zhaoyilunnn:
ZX-calculus has been a unified language describing fault-tolerrant quantum computation, and PyZX has been considered an input to the [TQEC](https://github.com/tqec/tqec/) framework.

To enable an end-to-end compilation flow and support various high-level quantum programming frameworks such as qiskit, cirq, ..., it is desirable to have a robust QASM parser since OpenQASM 2.0 is currently a widely used interface.

There have been some issues with loading a QASM program into PyZX, cf. https://github.com/tqec/tqec/issues/708

This PR (still work-in-progress) intends to take an initial step by supporting `reset` and `measurement` instructions in QASM.

## Test Program

<details> <summary> click to expand </summary> 

```python
import random
import pyzx as zx
from pyzx import Circuit

import sys


# qasm_file = sys.argv[1] if len(sys.argv) > 1 else "steane_encoding.qasm"

qasm_str = """
OPENQASM 2.0;
include "qelib1.inc";
qreg ancilla[3];
qreg data[7];
creg bits[3];
h ancilla[0];
cx ancilla[0],data[0];
cx ancilla[0],data[1];
cx ancilla[0],data[2];
cx ancilla[0],data[3];
h ancilla[0];
measure ancilla[0] -> bits[0];
h ancilla[1];
cx ancilla[1],data[0];
cx ancilla[1],data[1];
cx ancilla[1],data[4];
cx ancilla[1],data[5];
h ancilla[1];
measure ancilla[1] -> bits[1];
h ancilla[2];
cx ancilla[2],data[0];
cx ancilla[2],data[2];
cx ancilla[2],data[4];
cx ancilla[2],data[6];
h ancilla[2];
measure ancilla[2] -> bits[2];
"""

# circuit = Circuit.from_qasm_file(qasm_file)
circuit = Circuit.from_qasm(qasm_str)

# g = circuit.to_graph(compress_rows=False)
# g = circuit.to_graph(compress_rows=True, fix_io=True)
g = circuit.to_graph()

zx.full_reduce(g)
zx.to_rg(g)
random.seed(42)
fig_data = zx.draw(g, labels=True)

```

</details>

## Measurements

Unlike `reset` which is not yet recognized by the current parser, `measure` is recognized and a `Measurement` gate is created in `Circuit`. However, currently the above test program raises

```
Traceback (most recent call last):
  File "/home/zhaoyilun/forks/pyzx/temp/test_qasm_load.py", line 44, in <module>
    g = circuit.to_graph()
        ^^^^^^^^^^^^^^^^^^
  File "/home/zhaoyilun/forks/pyzx/pyzx/circuit/__init__.py", line 268, in to_graph
    return circuit_to_graph(self if zh else self.to_basic_gates(),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zhaoyilun/forks/pyzx/pyzx/circuit/graphparser.py", line 139, in circuit_to_graph
    gate.to_graph(g, q_mapper, c_mapper)
  File "/home/zhaoyilun/forks/pyzx/pyzx/circuit/gates.py", line 1237, in to_graph
    DiscardBit(self.result_bit).to_graph(g, q_mapper, c_mapper)
  File "/home/zhaoyilun/forks/pyzx/pyzx/circuit/gates.py", line 1195, in to_graph
    r = c_mapper.next_row(self.target)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zhaoyilun/forks/pyzx/pyzx/circuit/gates.py", line 73, in next_row
    return self._rows[l]
           ~~~~~~~~~~^^^
KeyError: 0
```

### Root Cause Analysis

The reason of the above error is that

1. During parsing, a `Circuit` is created with no classical bits: https://github.com/zxcalc/pyzx/blob/e89a336346a825794abd600b78df152ca5f27f0e/pyzx/circuit/qasmparser.py#L80
2. However, a `Measurement` is created with `result_register` and thus `result_bit`: https://github.com/zxcalc/pyzx/blob/e89a336346a825794abd600b78df152ca5f27f0e/pyzx/circuit/qasmparser.py#L152
3. When there exists a  `result_bit`, a `DiscardBit` node will be created when transforming to a graph: https://github.com/zxcalc/pyzx/blob/e89a336346a825794abd600b78df152ca5f27f0e/pyzx/circuit/gates.py#L1234-L1237, which assumes the `result_bit` is a key in the `_rows` of `c_mapper`.
4. However, since the `Circuit` has no classical bits, it will not go to this branch: https://github.com/zxcalc/pyzx/blob/e89a336346a825794abd600b78df152ca5f27f0e/pyzx/circuit/graphparser.py#L104-L109, thus `_rows` in `c_mapper` is empty, resulting in the `KeyError`

### Solutions

To fix this issue, I've considered two approaches

1. **The first is to support classical bits**. During parsing, support `creg` and create a `Circuit` with classical bits: https://github.com/Zhaoyilunnn/pyzx/blob/a5c216f1325b1ccea67e60bc1190b166accea513/pyzx/circuit/qasmparser.py#L185-L189, and add labels in `c_mapper`: https://github.com/Zhaoyilunnn/pyzx/blob/a5c216f1325b1ccea67e60bc1190b166accea513/pyzx/circuit/graphparser.py#L124-L147.  Here I tried two choices, 
 - One is creating new vertices for classical bits, the resulted graph is like:
    
image

- Another is "link" classical bits to existing vertices, resulting in a graph
    
image

2. **The second is to simply remove remove classical bits**. That is, create a `Measurement` with no `result_bit`. This gives the following graph which looks correct to me (produced by the [current state of this PR](https://github.com/Zhaoyilunnn/pyzx/commit/0a61c567ac6fa76c1cb4ad5137005284318e1410)). However, when dealing with classical control flow, e.g., feedforward operations, I think we do need classical bits. 

image

### Questions

How should a classical bit be represented in a ZX-diagram?
- Should we create a new BOUNDARY vertex for each classical bit?
- How to connect a classical bit with a measurement operation (a node with a "ground" symbol)?
- Should we distinguish between quantum wires and classical wires? I suspect that currently PyZX does not distinguish these two?

## Reset

To support reset, I have a few questions.
1. In a ZX-diagram, are they equivalent to a state initialization like the below?
image

2. Should we automatically reset/init each qubit when transforming a circuit to a graph? An OpenQASM program often implicitly assumes that all qubits are initialized to $|0\rangle$, however, without explicitly applying a state to inputs, these qubits are just boundary vertices. But applying explicit initializations may be required for some circuits to perform further re-write using ZX-calculus rules? I am not sure about this.


jvdwetering:
Hi Yilun, thanks for looking into this. I'm also not sure what the right behaviour is in this case. I think if there are no classically controlled operations that your solution 2 would be nice.
A potential solution is to just offer all options to the user with a sensible default. To translate to a Circuit is probably not ambiguous, so it is only the translation to a Graph where choices must be made. Potentially this could be done by setting properties of the Circuit beforehand? Something like `Circuit.set_classical_bit_representation`, and (for your other question) `Circuit.initialize_inputs_to_state`

Zhaoyilunnn:
Thanks for the suggestions, then I'll try to finalize this PR

> it is only the translation to a Graph where choices must be made. 

Got it, then I plan to add some options to the `to_graph()` interface of `Circuit`

---

Lastly, my editor has a formatter that runs automatically on-save, resulting in non-sense changes that just standarizes code format, should I avoid this to make it easier to reviewers? :)

pyzx/circuit/__init__.py
```diff
@@ -259,14 +259,14 @@ def from_graph(g:BaseGraph, split_phases:bool=True) -> 'Circuit':
         from .graphparser import graph_to_circuit
         return graph_to_circuit(g, split_phases=split_phases)
 
-    def to_graph(self, zh:bool=False, compress_rows:bool=True, backend:Optional[str]=None) -> BaseGraph:
+    def to_graph(self, zh:bool=False, compress_rows:bool=True, backend:Optional[str]=None, apply_state_effect:bool=False) -> BaseGraph:
```
jvdwetering: Add description of new parameter to docstring.

pyzx/circuit/__init__.py
```diff
@@ -259,14 +259,14 @@ def from_graph(g:BaseGraph, split_phases:bool=True) -> 'Circuit':
...
         """Turns the circuit into a ZX-Graph.
         If ``compress_rows`` is set, it tries to put single qubit gates on different qubits,
         on the same row."""
         from .graphparser import circuit_to_graph
 
         return circuit_to_graph(self if zh else self.to_basic_gates(),
-            compress_rows, backend)
+            compress_rows, backend, apply_state_effect)
```
jvdwetering: Add description of new parameter to docstring

pyzx/circuit/graphparser.py
```diff
@@ -86,15 +86,22 @@ def graph_to_circuit(g:BaseGraph[VT,ET], split_phases:bool=True) -> Circuit:
...
-def circuit_to_graph(c: Circuit, compress_rows:bool=True, backend:Optional[str]=None) -> BaseGraph[VT, ET]:
-    """Turns the circuit into a ZX-Graph.
+def circuit_to_graph(c: Circuit, compress_rows:bool=True, backend:Optional[str]=None, apply_state_effect: bool = False) -> BaseGraph[VT, ET]:
+    """
+    Turns the circuit into a ZX-Graph.
     If ``compress_rows`` is set, it tries to put single qubit gates on different qubits,
-    on the same row."""
+    on the same row.
+
+    If ``apply_state_effect`` is set, the graph will have fixed Z-basis inputs and outputs
```
jvdwetering: I don't really understand this comment. Is it that if `apply_state_effect==True` then we set all inputs and outputs to have |0> or |0< connected to it, producing a scalar. And if have it set to `False` then this only happens when we use Measurement?

Zhaoyilunnn: If `apply_state_effect==True`, then we set all inputs to have $|0\rangle$ connected to them. For outputs that have a `Measurement` gate connected, we do not represent the `Measurement` gate as a node with ground symbol, rather, we remove the `Measurement` and have $\langle 0|$ connected to the output. 

If we set it to False, we don't apply states or effects to the graph inputs / outputs, no matter we use `Measurement` or not.

Zhaoyilunnn:
> Thanks for the additions, I have a couple of questions

Thanks for reviewing this PR, I've updated the docstring.

pyzx/circuit/graphparser.py
lia-approves: My concern is that this hotfix only handles the edge case of circuits with both all inputs initialized to |0> and all measurements post-selected to <0|. In the absence of a (global with respect to needing to check all measurements in the entire OpenQASM input) check that this precondition is indeed satisfied, once `measure` in `QASMParser` is allowed as an OpenQASM instruction, the expected user behavior would be to assume that it can handle each measure individually, whether it's post-selected or not.

RazinShaikh:
As an alternative to using the ground symbol for measurements, we could also use symbolic Boolean phases, labelled by 'aπ' where the Boolean variable 'a' represents the measurement outcome. This is commonly done in ZX papers and PyZX supports these symbolic phases.

pyzx/circuit/graphparser.py
lia-approves: I think a related underlying issue is that postselection isn't an instruction in OpenQASM, because it was designed to treat simulating quantum circuits like running experiments.
In the [OpenQASM 2.0 spec](https://arxiv.org/pdf/1707.03429), the `measure` instruction just measures the specified qubit / quantum register, and writes the measurement outcome bit / bitstring to the specified bit / classical register. Postselecting would be to additionally enforce what value the measurement outcome is, which is external to what you can do in OpenQASM 1.0/2.0/3.0.

Since for many practical reasons people do want to allow both postselected and non-postselected measurements in the same circuit, to stick with OpenQASM 1.0/2.0/3.0, I think the best approach is to extend the OpenQASM `measure` instruction to take in an optional Boolean argument if each measurement is postselected or not (default being not), and if so another optional Boolean argument (default being 0) for what value it postselects to.

I should add, I don't think the behavior of `measurement` for QASMParser should ever be to discard (aka trace out, with the ground symbol). Discarding is equivalent to measuring and not observing the outcome, whereas with quantum circuits measuring generally is taken to mean a different linear map because you are observing the outcome.
I think the default behavior of parsing `measure` should be to initialize the 1-input, 0-output X spider with phase `a $\pi$`, where a is a Boolean value (as a symbolic variable in PyZX, which supports Sympy) equal to the measurement outcome. This has been the default representation of Z-basis measurement in ZX-calculus since the start; more recently we wrote this up formally as mixed-state quantum channels in bottom of pg 7 of [https://arxiv.org/pdf/2409.13541](https://arxiv.org/pdf/2409.13541).

lia-approves: I think this will work for your purposes, without needing to extend OpenQASM 2.0: after parsing all `measure` instructions with the default interpretation of `measure` in the ZX-calculus literature, we can simply add a function in PyZX that sets all symbolic Boolean variables in a ZX diagram to be 0, and you can simply call that after parsing.

lia-approves: Since the approach in the comment above would be slowed by unnecessarily invoking Sympy (albeit truest to expected behavior), I think the best solution is:
- Default behavior: Parse every measurement as `<a|` for Sympy bool `a`
- Optional bool vec param to the parser: for each input, is 1 if init to `|0>`, else 0; default is the all 0's vec
- Optional bool vec param to the parser: for each measurement (ordered by their appearance in the QASM file), is 1 if parse as `<0|`, else `<a|` for Sympy bool `a`; default is the all 0's vec
So you'd just have to call the parser with these 2 bool vec params each set to the all 1's vec.

Zhaoyilunnn: Thanks for the detailed instructions!

If I understand correctly, you mean that we should have two steps

1. g=circuit.to_graph()   In this step measure should be represented what it should be in ZX. 
2. g.some_transformation_fuction()   In this step we replace measures with 0 effects

If this is the case, I agree with this methodology, currently I add a apply_state_effect option to to_graph(), splitting to two steps looks way more elegant.


Regarding the first step, apparently I am not an expert in ZX, I will definitely implement the concensus method from ZX community. Previously the measure seems to have already been represented as ground symbol so I didn't change the behavior. Maybe I can understand this better during the incoming meeting

--- 

EDIT:

Based on your @lia-approves latest comments. I think the user experience should be

1. `circ = Circuit.from_qasm_file(some_boolean_variables=xxx)`
2. `g = circ.to_graph()`

This also looks good to me.  I think now the only debate here is how should measurement be represented in ZX, grounded symbol, or symbolic boolean phases?

Zhaoyilunnn: > Since the approach in the comment above would be slowed by unnecessarily invoking Sympy (albeit truest to expected behavior), I think the best solution is:
> 
> * Default behavior: Parse every measurement as `<a|` for Sympy bool `a`
> * Optional bool vec param to the parser: for each input, is 1 if init to `|0>`, else 0; default is the all 0's vec
> * Optional bool vec param to the parser: for each measurement (ordered by their appearance in the QASM file), is 1 if parse as `<0|`, else `<a|` for Sympy bool `a`; default is the all 0's vec
>   So you'd just have to call the parser with these 2 bool vec params each set to the all 1's vec.

~~If I understand correctly, it may be better and easier to add these boolean variables to the `to_graph` method of `Circuit`~~

---

Add them to the parser is also good to me

pyzx/circuit/__init__.py
```diff
@@ -259,14 +259,24 @@ def from_graph(g:BaseGraph, split_phases:bool=True) -> 'Circuit':
...
+        compress_rows:bool=True,
+        backend:Optional[str]=None,
+        init:Optional[List[bool]]=None,
+        post_select:Optional[List[int]]=None
+    ) -> BaseGraph:
         """Turns the circuit into a ZX-Graph.
         If ``compress_rows`` is set, it tries to put single qubit gates on different qubits,
-        on the same row."""
+        on the same row.
+        ``init`` denotes whether each input should be connected to |0\ranlge,
```
jvdwetering: typo in \ranlge
Also, you should be a bit more clear in what parameter does exactly. Like 'init[q] == True' means we set qubit q to be prepared in the |0> state, False means we leave it open. The default is open.
We could also make the argument be a bit more descriptive than 'init', maybe 'initialize_qubits' and 'postselect_qubits'?

pyzx/circuit/gates.py
```diff
@@ -1256,6 +1258,24 @@ def to_graph(self, g, q_mapper, c_mapper):
             g.add_edge((v,u), EdgeType.SIMPLE)
             c_mapper.set_next_row(self.result_bit, r+1)
 
+    def to_graph_symbolic_boolean(self, g, q_mapper):
+        """Represent the measurement as a node with symbolic boolean phases."""
+        r = q_mapper.next_row(self.target)
+        phase = new_var(name=f"a{self.target}", is_bool=True, registry=g.var_registry) 
+        _ = self.graph_add_node(g,
+            q_mapper,
+            VertexType.X,
```
dlyongemallo, line 1274: I just wanted to double-check that this is correct/intended. Isn't the default basis for measurement in OpenQASM the Z-basis? Or does an X-spider actually represent a Z-basis measurement, or did I miss an implicit Hadamard somewhere?

pyzx/circuit/__init__.py
```diff
@@ -259,14 +259,24 @@ def from_graph(g:BaseGraph, split_phases:bool=True) -> 'Circuit':
...
+        ``post_select`` denotes for each measurement whether it should be 
+        postselected to |0\rangle (0) or |1\ranlge (1)."""
```
dlyongemallo: Just wanted to check: Is `|0>` or `|1>` correct, or would the measurement actually postselect to `|+>` or `|->` with a `VertexType.X` as you have it?

pyzx/circuit/gates.py
```diff
@@ -1256,6 +1258,24 @@ def to_graph(self, g, q_mapper, c_mapper):
             g.add_edge((v,u), EdgeType.SIMPLE)
             c_mapper.set_next_row(self.result_bit, r+1)
 
+    def to_graph_symbolic_boolean(self, g, q_mapper):
+        """Represent the measurement as a node with symbolic boolean phases."""
+        r = q_mapper.next_row(self.target)
+        phase = new_var(name=f"a{self.target}", is_bool=True, registry=g.var_registry)
```
dlyongemallo: Maybe name the variables `m0`, etc., instead of `a0`?

pyzx/circuit/gates.py
RazinShaikh: It's unintuitive but the X spider represents a Z basis measurement because an X spider with phase 0 is |+> + |- > = √2 |0> and with phase π is |+> - |- > = √2 |1> .

RazinShaikh:
@jvdwetering I think this PR is ready to be merged :)

PR #351, opened 2025-09-30
Add documentation for symbolic.py

### LINKED ISSUE #290: symbolic.Var lacking documentation
Hello.

I am rewriting Symbolic.py to make the typeclass Poly generic, so that for example, one can represent Polynomials over finite fields, integers, etc. in a uniform way.  I am also adding support for fractions of polynomials as well.

I am almost done with modifying this file to this end, but I am running into problems with the Var class.  I am currently using Var as a wrapper for str, but there are other attributes of var which are not documented. For example, I have no idea what the following two attributes are meant to represent:

_is_bool: bool
_types_dict: Optional[Union[bool, Dict[str, bool]]]

Could someone please add some documentation, as I would really appreciate it.

--Cole

RazinShaikh:
Closes #290

### DIFF
--- a/pyzx/symbolic.py
+++ b/pyzx/symbolic.py
@@ -29,7 +29,7 @@ from fractions import Fraction
 
 
 class VarRegistry:
-    """Registry to track variable types for a specific graph"""
+    """Registry to track variable types (Boolean/continuous) for a specific graph"""
 
     types: Dict[str, bool]  # name -> is_bool
 
@@ -50,6 +50,13 @@ class VarRegistry:
 
 
 class Var:
+    """Symbolic variable that keeps track of its Boolean/continuous type.
+
+    Variables are always associated with a :class:`VarRegistry`, which records
+    whether the variable should be treated as Boolean (multiples of $\pi$) or
+    as a general real-valued parameter. Unless an explicit registry is
+    provided, a fresh registry is created so the variable can stand on its own.
+    """
     name: str
     _registry: VarRegistry
 
@@ -96,6 +103,9 @@ class Var:
         new_registry.set_type(self.name, self.is_bool)
 
 class Term:
+    """Product of symbolic variables with associated integer exponents.
+    Example: x^2 * y * z^3 is represented as Term([(x, 2), (y, 1), (z, 3)])
+    """
     vars: List[Tuple[Var, int]]
 
     def __init__(self, vars: List[Tuple[Var,int]]) -> None:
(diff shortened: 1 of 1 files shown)


pyzx/symbolic.py
```diff
@@ -50,6 +50,13 @@ def vars(self) -> Set[str]:
 
 
 class Var:
+    """Symbolic variable that keeps track of its Boolean/continuous type.
+
+    Variables are always associated with a :class:`VarRegistry`, which records
+    whether the variable should be treated as Boolean (multiples of $\pi$) or
```
jvdwetering: They can also be multiples of something else than pi right, like pi2?

pyzx/symbolic.py
```diff
@@ -138,8 +155,14 @@ def __lt__(self, other: 'Term') -> bool:
...
-        """Substitute variables in the term with the given values. Returns a tuple
-        of the coefficient and the new term.
+        """Evaluate variables present in ``var_map`` and return residual term.
+
+        The method extracts the numerical coefficient contributed by
+        substituted variables and returns both the coefficient and a new term
+        containing only the untouched variables. This is a building block for
+        partial evaluation of polynomials.
+
+        Example: substituting {x: 2} in x^2*y gives (4, y).
```
jvdwetering: I think I understand the example, but it doesn't match the syntax of how a Term is represented.

pyzx/symbolic.py
```diff
@@ -151,12 +174,18 @@ def substitute(self, var_map: Dict[Var, Union[float, complex, 'Fraction']]) -> T
...
 
     def rebind_variables_to_registry(self, new_registry: VarRegistry) -> None:
-        """Rebind all variables in this term to the given registry"""
+        """Rebind all variables in this term to the given registry."""
         for var, _ in self.vars:
             var.rebind_to_registry(new_registry)
 
 
 class Poly:
+    """A multivariate polynomial over integer, floating point, fractional, or complex coefficients.
```
jvdwetering: Do we want to say here that these are meant as phase variables, and hence that we should treat everything as being multiplied by pi?

pyzx/symbolic.py
```diff
@@ -364,6 +404,7 @@ def new_const(coeff: Union[int, Fraction]) -> Poly:
     maybe_placeholders=True)
 
 class PolyTransformer(Transformer):
+    """Lark transformer that builds :class:`Poly` instances from parse trees."""
```
jvdwetering: Maybe explain as well where we use this.

RazinShaikh:
Thanks, I have addressed these comments now

PR #358, opened 2025-10-13
Add support for fault-equivalent rewrite rules

boldar99:
Implements several fault-equivalent (FE) rewrite rules, including:
 - `fuse_1_FE`
 - `unfuse_1_FE`
 - `unfuse_4_FE`
 - `unfuse_5_FE`
 - `unfuse_2n_FE`
 - `unfuse_2n_plus_FE`
 - `recursive_unfuse_FE`
Moreover, checkers for these rewrites and unsafe versions. 

Introduces a new `pyzx.ft_simplify` sub-module where `Rewrite` instances are initiated for each FE rewrite.


jvdwetering:
Thanks, that looks great!
I don't think this is feasible for this PR, but I'm putting this down for maybe the future: in the same way that we can verify the correctness of regular rewrites using tests based on calculating the underlying tensor, it would be nice if we can validate the FTness of these rewrites using some method

PR #360, opened 2025-10-17
Add fault edge

y-richie-y:
In relation to the discussion in #358, I've added a very basic implementation of fault edges in pyzx which doesn't allow any rewrites. This enables to open a corresponding PR in zxlive.

### DIFF
--- a/pyzx/editor_actions.py
+++ b/pyzx/editor_actions.py
@@ -128,7 +128,12 @@ def pauli_push(g: BaseGraph[VT,ET],
             rem_edges.append(edge)
             w2 = g.add_vertex(t,q,r,p)
             etab[upair(v,w2)] = [1,0]
-            etab[upair(n,w2)] = [1,0] if et == EdgeType.SIMPLE else [0,1]
+            if et == EdgeType.SIMPLE:
+                etab[upair(n,w2)] = [1,0]
+            elif et == EdgeType.HADAMARD:
+                etab[upair(n,w2)] = [0,1]
+            else:
+                raise ValueError(f"Cannot apply Pauli commutation through {et} edge")
             new_verts.append(w2)
         if not vertex_is_zx(g.type(v)): # v is H_BOX
             if len(new_verts) == 2:
--- a/pyzx/graph/multigraph.py
+++ b/pyzx/graph/multigraph.py
@@ -42,7 +42,7 @@ class Edge:
         self.h += h
         self.w_io += w_io
         self.fault_edge += fault_edge
-        if self.s < 0 or self.h < 0:
+        if self.s < 0 or self.h < 0 or self.w_io < 0 or self.fault_edge < 0:
             raise ValueError('Cannot have negative edges')
         if self.w_io not in (0,1):
             raise ValueError('Invalid number of W-IO edges')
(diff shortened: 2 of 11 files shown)


pyzx/rewrite_rules/basicrules.py
```diff
@@ -195,6 +195,8 @@ def fuse(g: BaseGraph[VT,ET], v1: VT, v2: VT) -> bool:
             other_vertex = v1
         if other_vertex == v1 and g.edge_type(e) == EdgeType.SIMPLE:
             continue
+        if other_vertex == v1 and g.edge_type(e) == EdgeType.FAULT_EDGE:
```
RazinShaikh: Just to make sure I understand this line, when fusing two spiders that are connected by a simple edge, if there is a faulty edge between them, this condition prevents the fusion from creating a fault edge self-loop. Is this the intended behaviour?

y-richie-y: Yeah thats not right. Good spot!

pyzx/graph/multigraph.py
RazinShaikh: This method removes w_io edge even if you want to remove fault_edge

RazinShaikh (changes requested):
There are still several places in the code that needs to be updated. You should do a find for "EdgeType." and update the relevant methods.

pyzx/graph/multigraph.py
y-richie-y: Good catch! I've added more checks here and other parts of the code that assumed non-simple edges are Hadamard.

RazinShaikh:
The edge annotations can go on these fault-edges as edge_data. The rewrites on the fault-edges can be defined according to the noise models that interpret the edge data appropriately. This approach avoids matching rewrites for ideal (aka simple) edges regardless of the noise model.

pyzx/editor_actions.py
```diff
@@ -132,15 +128,20 @@ def pauli_push(g: BaseGraph[VT,ET],
             rem_edges.append(edge)
             w2 = g.add_vertex(t,q,r,p)
             etab[upair(v,w2)] = [1,0]
-            etab[upair(n,w2)] = [1,0] if et == EdgeType.SIMPLE else [0,1]
+            if et == EdgeType.SIMPLE:
+                etab[upair(n,w2)] = [1,0]
+            elif et == EdgeType.HADAMARD:
+                etab[upair(n,w2)] = [0,1]
+            else:
+                raise ValueError(f"Cannot apply Pauli commutation through {et} edge")
```
RazinShaikh, line 136: Raising the ValueError here breaks the rewrite. Consider the following diagram where we want to push the pauli. Since the pauli matcher has not been modified, the rewrite will be matched but when you apply the rewrite, it will fail. So you need to either add the correct edge type when pushing pauli or modify the pauli matcher to not match when fault edges are involved.
Image

pyzx/editor_actions.py
RazinShaikh, line 263: Applying the bialgebra rule changes the fault edge to a hadamard edge.
Image

Image

pyzx/rewrite_rules/editor_actions.py
RazinShaikh, line 55: Color change matches when one of the edge is a fault-edge but applying the rewrite fails due to an error caused by the toggle_edge method

pyzx/rewrite_rules/rules.py
RazinShaikh, line 248: "list index out of range" when the diagram has a fault-edge. I believe any time you see "etab", the rewrite is going to fail if the diagram has fault-edge/

RazinShaikh (changes requested):
I didn't check every rewrite rule but a lot of them are broken. I have given a few examples below. There are many places in the code that assumes that an edge is either a simple edge or a hadamard edge. So the rewrites involving fault-edge either behaves incorrectly or simply fail. Moreover, whenever you see "etab" that is likely to fail with fault-edge as that dictionary assumes the simple-hadamard duopoly. You probably need to check through all the methods and see if they work when the diagram has a fault edge.

PR #361, opened 2025-10-18
make the mypy versions consistent between the CI and test_requirements

dlyongemallo:


dlyongemallo:
The latest version of [`mypy`](https://pypi.org/project/mypy/) (1.18.2 at the moment), which is installed from `test_requirements.txt`), is stricter than the version used in the CI (1.5.1) and shows more errors. This PR makes the behaviour of the local development environment consistent with the CI.

(Conversely, it might be worthwhile to consider upgrading the version of `mypy` used for CI/development.)

PR #362, opened 2025-10-18
remove self-loops when `spider_simp` is called on multigraphs

### LINKED ISSUE #352: spider_simp on multigraphs doesn't remove self-loops
To reproduce:

```python
g = zx.graph.multigraph.Multigraph()
g.set_auto_simplify(False)
g.add_vertex(B, row=0)
g.add_vertex(X, row=1)
g.set_inputs([0])
g.add_edge((0,1),SE)
g.add_edge((1,1),SE)
zx.draw(g)
zx.spider_simp(g,quiet=True)
zx.draw(g)
```

The second picture still has the self-loop. This breaks other functions, e.g. `to_graph_like`.

dlyongemallo:
Fixes #352.


jvdwetering:
Thanks! It's great that you are looking at all of these bugs.

PR #363, opened 2025-10-19
in `dict_to_graph`, read the `inputs`, `outputs`, and `scalar` keys

### LINKED ISSUE #357: Json parser never loads the scalar into the graph
The `dict_to_graph` method ignores the scalar key so a graph loaded from json always has the empty scalar 1.

dlyongemallo:
Fixes #357.

### DIFF
--- a/pyzx/graph/scalar.py
+++ b/pyzx/graph/scalar.py
@@ -84,6 +84,20 @@ class Scalar(object):
     def __complex__(self) -> complex:
         return self.to_number()
 
+    def __eq__(self, other: object) -> bool:
+        """Compares all fields for equality.
+
+        Note that two Scalars can have the same output for to_number() even though not all their fields are equal."""
+        if not isinstance(other, Scalar):
+            return NotImplemented
+        return (self.power2 == other.power2 and
+                self.phase == other.phase and
+                self.phasenodes == other.phasenodes and
+                self.sum_of_phases == other.sum_of_phases and
+                self.floatfactor == other.floatfactor and
+                self.is_unknown == other.is_unknown and
+                self.is_zero == other.is_zero)
+
     def polar_str(self) -> str:
         """Returns a human-readable string of the scalar in polar format"""
         r,th = cmath.polar(self.to_number())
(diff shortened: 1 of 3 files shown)


pyzx/graph/scalar.py
```diff
@@ -84,6 +84,20 @@ def __str__(self) -> str:
     def __complex__(self) -> complex:
         return self.to_number()
 
+    def __eq__(self, other: object) -> bool:
+        """Compares all fields for equality.
+
+        Note that two Scalars can have the same output for to_number() even though not all their fields are equal."""
+        if not isinstance(other, Scalar):
+            return NotImplemented
```
jvdwetering, line 92: Should this be an exception, instead of returning something?

dlyongemallo: For `__eq__()`, `NotImplemented` has a special meaning ([reference](https://docs.python.org/3.11/library/constants.html#NotImplemented)).

In particular, `scalar == non_scalar` first calls `scalar.__eq__(non_scalar)`, and if it returns `NotImplemented`, then `non_scalar.__eq__(scalar)` is called (which may be implemented). Here, we declare that `Scalar` has no idea how to check for equality with something which is not a `Scalar`, but some non-`Scalar` class may well have some rule for how to compare itself to a `Scalar`.

So this is the correct behaviour.

jvdwetering: Ah, very interesting! Thanks!

RazinShaikh:
It looks like you merged the main branch into this branch instead of the opposite.

tests/test_jsonparser.py
```diff
@@ -347,15 +370,5 @@ def test_zbox_label_roundtrip(self):
...
-        js = json.dumps(d)
-        d2 = json.loads(js)
-        g3 = dict_to_graph(d2)
-
-        self.assertEqual(g3.vdata(v1, 'label'), Fraction(3, 4))
-        self.assertEqual(g3.vdata(v2, 'label'), poly_label)
-        self.assertEqual(g3.vdata(v3, 'label'), 2.5 + 1.3j)
-        self.assertEqual(g3.vdata(v1, 'name'), 'my vertex')
-        self.assertEqual(g3.vdata(v2, 'custom'), 42)
-
```
dlyongemallo, line 359: The merge seems to have somehow deleted these lines (which are unrelated to this MR). I'll restore them.

PR #365, opened 2025-11-02
Added double-wire + linear combination

SarthakTandon9:


jvdwetering:
Hi Sarthak, can you say a bit more about what this does and what you can use it for? I would also suggest adding a Jupyter Notebook to demonstrate the functionality.

Issue #367, opened 2025-11-10
tikz_to_graph method doesn't work with symbolic phases


RazinShaikh:
The `tikz_to_graph` method raises an error when it encounters a symbolic phase as it can't parse it as a fraction. In that case, it should actually use the `string_to_phase` method in jsonparser.py which can correctly deal with symbolic phases and create a `Poly` phase.

PR #369, opened 2025-11-17
String to phase function added in the editor

DenisPoisson:
* Fixes a failure in `tikz_to_graph` when encountering symbolic (non-fraction) phases.
* Adds a fallback: if fractional phase parsing fails, the code now uses `string_to_phase`.
* Ensures TikZ diagrams with symbolic phases (e.g., `pi/4`, `a`, `theta`) load correctly.
* Improves robustness without modifying existing `editor.py` logic unnecessarily.
* Keeps backward compatibility and avoids breaking existing TikZ parsing behavior.

### DIFF
--- a/pyzx/editor.py
+++ b/pyzx/editor.py
@@ -31,7 +31,6 @@ from .graph import Scalar
 from .graph.graph import GraphS
 from .rewrite_rules import rules
 from . import tikz
-from graph.jsonparser import string_to_phase
 
 from .editor_actions import MATCHES_VERTICES, MATCHES_EDGES, operations, operations_to_js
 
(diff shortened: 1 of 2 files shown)


RazinShaikh:
Sorry, I believe you misunderstood the issue #367. You shouldn't change anything in editor.py for that issue. There is some code inside tikz_to_graph which parses the phase from the string to a fractional phase. But when the phase is not a fraction, it fails. When it fails to parse the phase, it should use the existing string_to_phase method from jsonparser.py

DenisPoisson:
Thank you for the clarification! That makes the issue much clearer. I’ll
revisit the approach and open a new PR with the correct fix using
string_to_phase inside tikz_to_graph. Appreciate your guidance — I’ll get
back to you soon with the updated PR.

RazinShaikh:
Hi, I think the PR mostly looks good to me. However, this PR changes several lines that are not directly relevant to the issue, including the deleting/modifying comments in the code and adding/deleting new lines where not necessary. I believe these are mostly stylistic and no functionality is changed but it would be cleaner to only change things that are directly about the issue. If you want to improve code formatting/quality, it might be better to do it in a different PR.

PR #372, opened 2025-12-13
Add Pauli web computation for all Clifford diagrams

maximilianruesch:
Adds relevant methods to PyZX to calculate Pauli webs of arbitrary Clifford diagrams. The method is based on [Borghan's Thesis](https://www.cs.ox.ac.uk/people/aleks.kissinger/papers/borghans-thesis.pdf) and was adapted as a byproduct of [arXiv:2510.08477](https://arxiv.org/abs/2510.08477).

The implementation is directly derived from https://github.com/maximilianruesch/fault-gadgets/tree/develop/src/faultgadget/web. As a consideration, one may want to adapt the `PauliWeb` implementation in that repo as well, since it is considerably slimmer than the existing one, but also has slightly different behavior (i.e. it does not know about the underlying graph, so it cannot auto-convert given whole-edge Paulis to color-changing half-edge Paulis).

> [!IMPORTANT]
> Since the `pyzx.pauliweb` namespace was already occupied, this module lives in `pyzx.web`. One may decide to change this, and one should probably decide how to deal with the old implementation anyway.

> [!WARNING]
> Since this repository uses merge commits, this branch may be cleaned / force pushed at will before merging.


jvdwetering:
Hi Max,

Thanks for the work! Just to confirm, this adds to the current pauliweb.py the ability to calculate the internal pauliwebs, and not just the ones involving inputs and outputs?

There's a couple of methods you introduce that don't have docstrings, could you still add those?

maximilianruesch:
Hi John!

Yes, this adds the ability to calculate internal Pauli webs (herein called "detecting regions", as opposed to the non-internal webs called "stabilisers" / "stabilising webs"). These are exposed via convenience methods individually, or can be calculated together (see `compute.py`).

However, this is not something to replace `pauliweb.py` entirely, since that module still provides information that is currently inaccessible from the new module, namely time ordering obtained through gflow. However, this new module can generate Pauli webs for all Clifford diagrams, as opposed to just those for which a gflow could be obtained.

I will add the docstrings soon!

pyzx/web/compute.py
```diff
@@ -0,0 +1,115 @@
...
+    determine_ordering,
+    create_firing_verification,
+    convert_firing_assignment_to_web_prototype,
+)
+from ..graph.base import BaseGraph
+from ..pauliweb import PauliWeb
+
+
+def _compute(
+    graph: BaseGraph[int, Tuple[int, int]], *, stabilisers: bool, detecting_regions: bool
```
jvdwetering, line 32: What is this '*' here?

maximilianruesch: It effectively swallows all arguments passed by position, and thus makes all arguments that come after it keyword-only. A neat trick to make boolean parameters have the description of what they do at their callsite, which is a good practice in general and even more so for those parameters that are called with literals (IDEs often provide inline hints to mitigate this, but this makes it clearer for those that do not, and code viewers such as the one on GitHub).

Not doing this may lead to the "boolean trap" and can be considered an antipattern. The ruff linter even has a lint for it (which it derived from a flake8 package): https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/

jvdwetering: Oh right, I've read about this feature when it was introduced in Python, but I had never used it myself. Very nice.

maximilianruesch:
> Am I also correct in understanding that the difference between your web implementation and pauliweb.py is that the latter can handle Clifford+T, but only in the unitary case, and your implementation only handles Cliffords, but can handle arbitrary post-selected diagrams?

Yes, concerning my implementation that is correct, concerning the `pauliweb.py` I am fairly sure that you are correct. It can handle Clifford+T since that is what I have observed the `tqec` group do at some point, but it requires gflow and thus (as far as I know) unitarity.

>Do you think it makes sense to have some overarching function that makes this selection for the user, so that they don't need to know this distinction?

It might make sense to add it for some users, but since the performance characteristics of the two approaches might be vastly different and users might want an exception to be raised whenever their diagram is not Clifford / not unitary (whichever way round the implementation they target does not support), I think it is safer for users to handle this explicitly.

> Also, I might have missed it, but do you raise an exception if a user tries to create a Pauliweb of a non-Clifford diagram?

Yes, during conversion of the graph to red-green form: https://github.com/maximilianruesch/pyzx/blob/c40644854e391401dd00a4d11d0a121be8ee3cd5/pyzx/web/red_green.py#L227-L232

PR #388, opened 2026-01-29
Add H-box cancellation rewrite rule

### LINKED ISSUE #200: Hadamard cancellation rule
Unless I am missing it there is no "Hadamard cancellation" implemented. 

Consider the following graph: 

```
g = zx.Graph() 

v0 = g.add_vertex(zx.VertexType.BOUNDARY, 0, 0)
v1 = g.add_vertex(zx.VertexType.H_BOX, 0, 1)
g.add_edge((v0, v1))
v2 = g.add_vertex(zx.VertexType.H_BOX, 0, 2)
g.add_edge((v1, v2))
v3 = g.add_vertex(zx.VertexType.BOUNDARY, 0, 3)
g.add_edge((v2, v3))
```

``full_reduce``, ``clifford_simp`` have no effect on it (nor did I find any other API that cancels the two Hadamards.

dlyongemallo:
Fixes #200.

### DIFF
--- a/pyzx/hsimplify.py
+++ b/pyzx/hsimplify.py
@@ -145,6 +145,9 @@ had_edge_to_hbox_simp = RewriteSimpDoubleVertex(check_hadamard_edge, unsafe_had_
 hbox_to_had_edge_simp = RewriteSimpSingleVertex(check_hadamard, unsafe_replace_hadamard)
 """Converts an h-box connecting the given vertices into a hadamard edge. Can be run automatically on the entire graph."""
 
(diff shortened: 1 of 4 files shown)


jvdwetering:
Thanks for implementing! Currently `simplify.py` contains the rewrites involving ZX-diagrams while `hsimplify.py` contains the rewrites that involve H-boxes. I'm not entirely sure what would be best to do though, since cancelling two Hadamards is a pretty fundamental rule that also plays a role in vanilla ZX. It's just that pyzx usually always treats Hadamards as a type of edge.
Maybe a solution is to define it in hsimplify, but then import it into simplify so that it can also be found in a place where people would expect it?

PR #389, opened 2026-01-29
Add add_phase_gadget() method to BaseGraph and a PhaseGadget gate frontend

### LINKED ISSUE #36: Feature Request: Adding Phase Gadgets
Is there an easy way of adding a phase gadget to a graph / circuit?

dlyongemallo:
Fixes #36.


dlyongemallo:
Yes, the difference is whether the graph conversion should create the phase gadget structure directly, or decompose to basic gates (CNOT, ZPhase). I agree it's confusing, but it would also be wrong to change the existing behaviour. I'll add a flag to control the composition, how's that?

PR #394, opened 2026-01-31
disallow phases with complex coefficients except for Z-box

### LINKED ISSUE #380: Incomplete handling of complex phases in .to_adjoint()
[/pyzx/circuit/gates.py Line 233](https://github.com/zxcalc/pyzx/blob/9dcf8444cc53821187f112742d6a5b696b01ec51/pyzx/circuit/gates.py#L233) sets all phases to their negative for the adjoint where presumably complex phases should be handled differently.

dlyongemallo:
Fixes #380.

### DIFF
--- a/pyzx/graph/base.py
+++ b/pyzx/graph/base.py
@@ -25,7 +25,7 @@ from typing_extensions import Literal, GenericMeta # type: ignore # https://gith
 import numpy as np
 
 from ..symbolic import Poly, VarRegistry
-from ..utils import EdgeType, VertexType, toggle_edge, vertex_is_zx
+from ..utils import EdgeType, VertexType, toggle_edge, vertex_is_zx, get_z_box_label, set_z_box_label
 from ..utils import FloatInt, FractionLike
 from ..tensor import tensorfy, tensor_to_matrix
 
@@ -504,6 +504,9 @@ class BaseGraph(Generic[VT, ET], metaclass=DocstringMeta):
             vtab[v] = i
             for k in self.vdata_keys(v):
                 g.set_vdata(i, k, self.vdata(v, k))
+            if adjoint and ty[v] == VertexType.Z_BOX:
+                label = get_z_box_label(self, v)
+                set_z_box_label(g, i, label.conjugate())
         for v in self.grounds():
             g.set_ground(vtab[v], True)
 
--- a/pyzx/graph/graph_ig.py
+++ b/pyzx/graph/graph_ig.py
@@ -21,6 +21,7 @@ except ImportError:
 	ig = None
 
 from .base import BaseGraph, VertexType, EdgeType
+from ..utils import assert_phase_real
 
 class GraphIG(BaseGraph):
 	"""Implementation of :class:`~graph.base.BaseGraph` using ``python-igraph`` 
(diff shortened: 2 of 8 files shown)


pyzx/circuit/gates.py
```diff
@@ -233,7 +233,11 @@ def copy(self: Tvar) -> Tvar:
     def to_adjoint(self: Tvar) -> Tvar:
         g = self.copy()
         if hasattr(g, "phase"):
-            g.phase = -g.phase
+            phase = g.phase
+            if isinstance(phase, Poly):
+                g.phase = (-phase).conjugate()
```
jvdwetering: Does this make sense? It looks like you are conjugating twice

pyzx/graph/scalar.py
```diff
@@ -115,17 +115,22 @@ def copy(self, conjugate: bool = False) -> 'Scalar':
         Returns:
             A copy of the Scalar
         """
+        def conjugate_phase(phase: FractionLike) -> FractionLike:
+            if isinstance(phase, Poly):
+                return (-phase).conjugate()
```
jvdwetering: Same here, it looks like you are conjugating twice

pyzx/circuit/gates.py
dlyongemallo: I don't understand. To compute the adjoint of `e^{i π phase}`, we need `e^{-i π phase*}` where `phase*` is the complex conjugate of `phase`. So `(-phase).conjugate()` is the correct value, isn't it?

Is the brackets that's the problem? Since negation and conjugation commute, I can remove them.

dlyongemallo:
I think I encountered a bug while trying to write a test for this. In [these lines](https://github.com/zxcalc/pyzx/blob/4301e5bb6e937ae50ae2ad141710e4d9e1bb92b4/pyzx/symbolic.py#L265-L266) from `symbolic.py`, terms with complex coefficients are filtered out when doing the `mod` operation. 

In `set_phase` it does a `phase % 2` [here](https://github.com/zxcalc/pyzx/blob/4301e5bb6e937ae50ae2ad141710e4d9e1bb92b4/pyzx/graph/graph_s.py#L330), terms with complex coefficients are all dropped. Surely this is wrong?

dlyongemallo:
Ah, that was noticed in #377 already, but for `multigraph.py`.

pyzx/graph/base.py
```diff
@@ -496,7 +495,8 @@ def copy(self, adjoint:bool=False, backend:Optional[str]=None) -> BaseGraph[VT,E
         maxr = self.depth()
         vtab = dict()
         for v in self.vertices():
-            i = g.add_vertex(ty[v],phase=mult*ph[v])
+            phase = ph[v] if not adjoint else adjoint_phase(ph[v])
```
RazinShaikh: if the vertex is a Z box, the adjoint should conjugate the value of the Z box label instead of changing the phase.

dlyongemallo: Done.

dlyongemallo:
> Btw, once the complex labeled H-box PR is merged, we should add a special case for the H box in the following.

Done in #405.

PR #395, opened 2026-01-31
Add rx and sx gate support to full_optimize

### LINKED ISSUE #215: pyzx unable to optimize when circuit contains rx or sx gate
Using the most recent version of pyzx (0.8.0), with the following code
```Python
import pyzx as zx
from pyzx.circuit import Circuit, CNOT
c = Circuit(1)
c.add_gate('XPhase', 0, phase=3/4)
zx.draw(c)
zx.full_optimize(c.to_basic_gates())
```
and it returns with an error
" Unknown gate XPhase(0,phase=0.75). Maybe simplify the gates with circuit.to_basic_gates()? "

dlyongemallo:
Fixes #215

### DIFF
--- a/pyzx/circuit/gates.py
+++ b/pyzx/circuit/gates.py
@@ -24,7 +24,7 @@ import math
 from fractions import Fraction
 from typing import Dict, List, Optional, Type, ClassVar, TypeVar, Generic, Set
 
-from ..utils import EdgeType, VertexType, FractionLike
+from ..utils import EdgeType, VertexType, FractionLike, settings
 from ..graph.base import BaseGraph, VT, ET
 from ..symbolic import new_var
 
@@ -449,6 +449,31 @@ class XPhase(Gate):
             return super().to_quipper()
         return 'QRot["exp(-i%X)",{!s}]({!s})'.format(math.pi*self.phase/2,self.target)
 
+    def to_basic_gates(self) -> List['Gate']:
+        # Handle symbolic (Poly) phases - return self unchanged.
+        if not isinstance(self.phase, (int, float, Fraction)):
+            return [self]
+
+        # Normalize phase to Fraction for consistent handling.
+        if isinstance(self.phase, float):
+            phase = Fraction(self.phase).limit_denominator(settings.float_to_fraction_max_denominator)
+        else:
+            phase = Fraction(self.phase)
+
+        # Normalize to [0, 2) range.
+        phase = phase % 2
+
+        # Identity gate - return empty list.
+        if phase == 0:
+            return []
+
+        # NOT/X gate - return as basic gate.
+        if phase == 1:
+            return [NOT(self.target)]
+
+        # General case: return normalized XPhase (keeps it as a basic gate).
+        return [XPhase(self.target, phase)]
+
     def tcount(self):
         return 1 if self.phase.denominator > 2 else 0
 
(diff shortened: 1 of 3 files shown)


pyzx/circuit/gates.py
```diff
@@ -449,6 +449,31 @@ def to_quipper(self):
...
         return 'QRot["exp(-i%X)",{!s}]({!s})'.format(math.pi*self.phase/2,self.target)
 
+    def to_basic_gates(self) -> List['Gate']:
+        # Handle symbolic (Poly) phases - return self unchanged.
+        if not isinstance(self.phase, (int, float, Fraction)):
+            return [self]
+
+        # Normalize phase to Fraction for consistent handling.
+        if isinstance(self.phase, float):
+            phase = Fraction(self.phase).limit_denominator(1024)
```
jvdwetering: This is a pretty rough denominator limit, I would increase the size of this, maybe 2**20 is a bit more reasonable?

pyzx/optimize.py
```diff
@@ -513,6 +513,23 @@ def parse_gate(self, g: Gate) -> None:
                 self.add_hadamard(c)
                 self.add_cnot(g)
 
+        elif isinstance(g, XPhase):
+            # XPhase(θ) = HAD·ZPhase(θ)·HAD, so we handle it by toggling hadamards and
+            # processing as a ZPhase.
+            toggle_element(self.hadamards, t)
+            # Normalize phase to Fraction if needed (skip for symbolic Poly phases).
+            from fractions import Fraction
```
jvdwetering: Import at top of file?

pyzx/optimize.py
```diff
@@ -513,6 +513,23 @@ def parse_gate(self, g: Gate) -> None:
...
+            if isinstance(g.phase, (int, float, Fraction)):
+                if isinstance(g.phase, float):
+                    phase = Fraction(g.phase).limit_denominator(1024)
```
jvdwetering: Make denominator larger?

jvdwetering (commented):
Main comment that I wonder whether it would be confusing to limit denominators to 1024. There are some quantum chemistry circuits I have seen where a higher resolution than that is needed. Ideally this would be some setting or parameter, but I think for now we can just be a bit more conservative and limit to something like 2**20 instead of 2**10.

pyzx/circuit/gates.py
dlyongemallo: I've made it a setting.

pyzx/circuit/gates.py
```diff
@@ -516,8 +541,27 @@ def __eq__(self, other: object) -> bool:
...
-    def to_basic_gates(self):
-        return [ZPhase(self.target, Fraction(1,2)), XPhase(self.target, -self.phase), ZPhase(self.target, -Fraction(1,2))]
+    def to_basic_gates(self) -> List['Gate']:
+        # Handle symbolic (Poly) phases - use original decomposition.
+        if not isinstance(self.phase, (int, float, Fraction)):
+            return [ZPhase(self.target, Fraction(1,2)), XPhase(self.target, -self.phase), ZPhase(self.target, -Fraction(1,2))]
+
+        # Normalize phase for the XPhase component.
+        if isinstance(self.phase, float):
+            x_phase = Fraction(-self.phase).limit_denominator(1024)
```
jvdwetering: Forgot to update here.

dlyongemallo: Good catch.

PR #396, opened 2026-01-31
Add basis choice for InitAncilla and PostSelect gates

### LINKED ISSUE #285: Add support for choosing basis in `InitAncilla` and `PostSelect`
Hi! I am currently implementing a lowering from Qualtran to pyzx (https://github.com/quantumlib/Qualtran/pull/1550).

I'm producing a `zx.Circuit`, and as qualtran natively supports states and effects, it would be very useful to have atomic zx gates that can represent these. I see that `InitAncilla` and `PostSelect` are there, but they add a `Z` spider, i.e. $\ket+$ and $\bra+$ respectively.

Could they be extended with flags to select the basis and basis state (i.e. one of $\ket0, \ket1, \ket+, \ket-$)? Or is there a more direct way to add states and effects?

dlyongemallo:
Fixes #285.

Also: Remove broken `PostSelect.to_graph` method. This is already correctly handled in `graphparser.py`.

### DIFF
--- a/pyzx/circuit/gates.py
+++ b/pyzx/circuit/gates.py
@@ -1210,6 +1210,12 @@ class InitAncilla(Gate):
         """Returns (VertexType, phase) for this ancilla state."""
         return self.STATE_MAP[self.state]
 
+    def reposition(self, mask, bit_mask=None):
+        g = self.copy()
+        g.label = mask[self.label]
+        g.target = g.label
+        return g
+
 
 class PostSelect(Gate):
     """Post-select a qubit in a specified state.
@@ -1245,6 +1251,13 @@ class PostSelect(Gate):
     def get_vertex_info(self):
         """Returns (VertexType, phase) for this post-selection state."""
         return self.STATE_MAP[self.state]
+
+    def reposition(self, mask, bit_mask=None):
+        g = self.copy()
+        g.label = mask[self.label]
+        g.target = g.label
+        return g
+
 
 class DiscardBit(Gate):
     name = 'DiscardBit'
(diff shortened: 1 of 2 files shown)


pyzx/circuit/gates.py
```diff
@@ -1167,20 +1167,97 @@ def to_graph(self, g, q_mapper, c_mapper):
...
+    Args:
+        label: The qubit label/index.
+        state: The initial state. One of '+', '-', '0', '1' for the
+            respective kets |+⟩, |-⟩, |0⟩, |1⟩. Defaults to '+'.
+    """
     name = 'InitAncilla'
-    def __init__(self, label):
+    STATE_MAP = _state_map
+
+    def __init__(self, label: int, state: str = '+'):
```
jvdwetering, line 1189: A more reasonable default for InitAncilla is the '0' state, as when a qc paper talks about initializing ancilla would specifying how to do it, it should be a 0 state.

pyzx/circuit/gates.py
```diff
@@ -1167,20 +1167,97 @@ def to_graph(self, g, q_mapper, c_mapper):
...
+    Args:
+        label: The qubit label/index.
+        state: The post-selection state. One of '+', '-', '0', '1' for the
+            respective bras ⟨+|, ⟨-|, ⟨0|, ⟨1|. Defaults to '+'.
+    """
     name = 'PostSelect'
-    def __init__(self, label):
+    STATE_MAP = _state_map
+
+    def __init__(self, label: int, state: str = '+'):
```
jvdwetering, line 1231: Same here for default value of PostSelect

PR #400, opened 2026-02-02
Add support for H-boxes with arbitrary complex labels

### LINKED ISSUE #167: Support H boxes with arbitrary complex labels
Similar to the complex labels of the Z box that are now supported

dlyongemallo:
Fixes #167.

### DIFF
--- a/pyzx/rewrite_rules/bialgebra_rule.py
+++ b/pyzx/rewrite_rules/bialgebra_rule.py
@@ -37,7 +37,8 @@ __all__ = ['check_bialgebra_reduce',
 
 from collections import defaultdict
 from typing import Callable, Optional, List, Tuple, Dict
-from pyzx.utils import EdgeType, VertexType, is_pauli
+from pyzx.utils import (EdgeType, VertexType, is_pauli,
+                        hbox_has_complex_label, get_h_box_label, set_h_box_label)
 from pyzx.graph.base import BaseGraph, VT, ET, upair
 
 RewriteOutputType = Tuple[Dict[Tuple[VT,VT],List[int]], List[VT], List[ET], bool]
--- a/pyzx/drawing.py
+++ b/pyzx/drawing.py
@@ -38,7 +38,7 @@ patches: Any = None
 lines: Any = None
 
 
-from .utils import settings, get_mode, phase_to_s, EdgeType, VertexType, FloatInt, get_z_box_label
+from .utils import settings, get_mode, phase_to_s, VertexType, FloatInt, get_z_box_label, get_h_box_label, hbox_has_complex_label
 from .graph.base import BaseGraph, VT, ET
 from .circuit import Circuit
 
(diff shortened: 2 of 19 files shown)


pyzx/rewrite_rules/bialgebra_rule.py
```diff
@@ -94,6 +95,9 @@ def unsafe_bialgebra(g: BaseGraph[VT,ET], v1: VT, v2: VT ) -> bool:
                 r = 0.4*g.row(other_vertex) + 0.6*g.row(v[i])
                 newv = g.add_vertex(g.type(v[j]), qubit=q, row=r)
                 g.set_phase(newv, g.phase(v[j]))
+                # Copy complex label if H-box has one.
+                if g.type(v[j]) == VertexType.H_BOX and hbox_has_complex_label(g, v[j]):
```
jvdwetering: v[i] and v[j] are always Z and X spiders here, and there phase I believe is always a Pauli, so this case here never applies.

dlyongemallo: What about the check on line 121 below? I think it has the same applicability as this check here, so if this one is removed, that one should as well, for consistency.

pyzx/rewrite_rules/bialgebra_rule.py
dlyongemallo, line 118: I am referring to this check for `H_BOX`. 

Based on the git history, it was added in commit 9dacf64c to `editor_actions.py`, before this was refactored here in commit ef23f18a. In [this line](https://github.com/zxcalc/pyzx/commit/ef23f18a#diff-122fd5a0bb584188884b772a1459d9af30ae4c77f7fcb420a37b60099b34bdceL252) there was originally a check for X-H algebra as well. 

So the code is inconsistent right now. `check_bialgebra` only allows Z-X pairs, but `unsafe_bialegra` (partially) handles H_BOX scalars. Should I remove both references (the existing one and the one I added) to H_BOX in `unsafe_bialgebra`? Or was the removal of X-H from `check_bialgebra` during the refactoring changes an accidental oversight?

dlyongemallo: @lara-madison If I am reading the git history right, you made the original refactoring. Maybe you can comment on this or know something about it?

RazinShaikh:
No worries, it can be done in a separate PR

pyzx/rewrite_rules/bialgebra_rule.py
dlyongemallo: To move this along, I'll remove this check here, but leave the other one (which already existed before this change) alone.

RazinShaikh:
I believe they are equivalent but I'm not sure which one is better to add in pyzx. Intuitively, the first version reduces from 2 H boxes to 1 so it looks more like a simplification. But maybe @jvdwetering has a better reasoning to decide which one to implement.

jvdwetering:
Yes, the first one is better. The only reason for the second one is because it fit into the restrictions of that paper (which don't exist in the pyzx implementation).
But also that should be a separate PR I think.

dlyongemallo:
@RazinShaikh @jvdwetering Another clarifying question: does the rule also apply if the two H-boxes are connected to an X-spider via Hadamard edges?

pyzx/rewrite_rules/bialgebra_rule.py
dlyongemallo: After looking through the git history, it seems that there was some attempt earlier to add X-H bialgebra support, but this was incomplete as pieces were still missing. Those pieces are mostly there now, so the feature can actually be implemented. I will follow this up in #418.

PR #407, opened 2026-02-21
add H-box average rule

dlyongemallo:
Implement rule (A) from arXiv:1805.02175: when two H-boxes sharing the same Z-spider neighbours are connected through a NOT gate, replace them with a single H-box whose label is the average of the two.

(Additionally, this also works with two H-boxes sharing the same X-spider neighbours connected through a NOT gate with Hadamard edges.)

### DIFF
--- a/tests/test_hbox.py
+++ b/tests/test_hbox.py
@@ -582,6 +582,179 @@ class TestParHboxAvgRule(unittest.TestCase):
         t_after = g.to_tensor(preserve_scalar=True)
         self.assertTrue(compare_tensors(t_before, t_after, preserve_scalar=True))
 
+    def test_no_match_z_hadamard_neighbors(self):
+        """Test that the average rule does not match Z-spider neighbours via Hadamard edges.
+
+        The average rule identity only holds for the canonical edge types
+        (Z+simple, X+Hadamard). Z+Hadamard produces a different tensor.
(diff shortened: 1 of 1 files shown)


dlyongemallo:
@RazinShaikh from [this comment](https://github.com/zxcalc/pyzx/pull/400#issuecomment-3945576796):
> I realized just now that there are more instances when this can apply. In general, if we have the top and the bottom spiders of the same color and the edges connected to the H boxes are both simple or both hadamard, then we can apply the rewrite. Moreover, if the top and the bottom spiders are of different colors, then we can still apply the rewrite if one of the two edges of both H boxes is a hadamard edge

IIUC this applies to the "intro" and "multiply" rules as well. The matchers for _all three_ rules need to handle all of these variants. Is that right?

dlyongemallo:
> I realized just now that there are more instances when this can apply. In general, if we have the top and the bottom spiders of the same color and the edges connected to the H boxes are both simple or both hadamard, then we can apply the rewrite. Moreover, if the top and the bottom spiders are of different colors, then we can still apply the rewrite if one of the two edges of both H boxes is a hadamard edge

Now I'm confused about whether this is actually true or not. I tested these cases and they don't appear to be valid after all. @RazinShaikh Can you take a look?

PR #412, opened 2026-03-01
TikZ metadata comment and symbolic phase round-trip (fixes #367)

### LINKED ISSUE #367: tikz_to_graph method doesn't work with symbolic phases

Bumsparkle:
Fixes #367

### DIFF
--- a/pyzx/tikz.py
+++ b/pyzx/tikz.py
@@ -295,7 +295,7 @@ def _normalise_tikzit_phase_expr(expr: str) -> Optional[str]:
         normalised_suffix = _normalise_tikzit_phase_expr(suffix)
         if normalised_suffix is None:
             return None
-        return f"({normalised_fraction})*({normalised_suffix})"
+        return f"({normalised_fraction})*{normalised_suffix}"
 
     tokens: List[str] = []
     i = 0
@@ -311,7 +311,24 @@ def _normalise_tikzit_phase_expr(expr: str) -> Optional[str]:
             command = expr[i + 1:j]
             if not command or command == "frac":
                 return None
-            tokens.append("pi" if command == "pi" else command)
+            token = "pi" if command == "pi" else command
+            # Keep simple LaTeX subscripts attached to commands, e.g. \alpha_1.
+            if j < len(expr) and expr[j] == "_":
+                k = j + 1
+                if k < len(expr) and expr[k] == "{":
+                    subscript, next_k = _extract_braced_group(expr, k)
+                    if subscript is None:
+                        return None
+                    token += "_" + subscript.strip()
+                    j = next_k
+                else:
+                    while k < len(expr) and (expr[k].isalnum() or expr[k] in "_[]"):
+                        k += 1
+                    if k == j + 1:
+                        return None
+                    token += expr[j:k]
+                    j = k
+            tokens.append(token)
             i = j
             continue
         if c in "{}":
@@ -320,6 +337,10 @@ def _normalise_tikzit_phase_expr(expr: str) -> Optional[str]:
             continue
         if c in "()+-*/^":
             tokens.append(c)
+            i += 1
+            continue
+        if c == "⋅":
+            tokens.append("*")
             i += 1
             continue
         if c.isdigit():
--- a/pyzx/tikz.py
+++ b/pyzx/tikz.py
@@ -281,35 +281,6 @@ def _extract_tikz_metadata(s: str) -> Tuple[str, Dict[str, bool]]:
     return clean_tikz, {}
 
 
-def _restore_symbolic_tikz_phases(
-    g: BaseGraph[VT, ET],
-    index_dict: Dict[int, VT],
-    node_lines: List[Tuple[int, str, str]],
-) -> None:
-    """Set phases for nodes whose label is symbolic (e.g. 'ff') that the main parser skipped."""
-    style_to_type: Dict[str, VertexType] = {
-        "z phase dot": VertexType.Z,
-        "x phase dot": VertexType.X,
-        "z dot": VertexType.Z,
-        "x dot": VertexType.X,
-    }
-    for vid, style, label in node_lines:
-        if style_to_type.get(style) not in (VertexType.Z, VertexType.X):
-            continue
-        if not any(ch.isalpha() for ch in label):
-            continue
-        if vid not in index_dict:
-            continue
-        v = index_dict[vid]
-        if v not in g.vertices():
-            continue
-        try:
-            phase = string_to_phase(label, g)
-            g.set_phase(v, phase)
-        except Exception:
-            continue
-
-
 def tikz_to_graph(
     s: str,
     warn_overlap:bool= True,
@@ -350,10 +321,12 @@ def tikz_to_graph(
         raise ValueError(tikz_error_message)
 
     g = Graph(backend)
+    if variable_types and hasattr(g, "var_registry"):
+        for name, is_bool in variable_types.items():
+            g.var_registry.set_type(name, is_bool)
     index_dict: Dict[int,VT] = {} # type: ignore
     position_dict: Dict[str,List[int]] = {}
     none_style_vertices: Set[VT] = set()  # type: ignore
-    node_lines_for_phase: List[Tuple[int, str, str]] = []  # (vid, style, label)
     for c,l in enumerate(lines[2:]):
         if l == r'\end{pgfonlayer}': break
         # l should look like
@@ -378,8 +351,6 @@ def tikz_to_graph(
             if ignore_parse_errors:
                 continue
             raise ValueError("Failed to parse node definition '%s': %s" % (orig_line, str(ex)))
-
-        node_lines_for_phase.append((vid, style.lower(), label))
 
         ty: VertexType
         is_none_style = False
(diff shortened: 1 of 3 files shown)


pyzx/tikz.py
```diff
@@ -220,6 +246,70 @@ def tikzit(g: Union[BaseGraph[VT,ET],Circuit,str], draw_scalar:bool=False) -> No
...
+        if vid not in index_dict:
+            continue
+        v = index_dict[vid]
+        if v not in g.vertices():
+            continue
+        try:
+            phase = string_to_phase(label, g)
+            g.set_phase(v, phase)
+        except Exception:
+            continue
```
RazinShaikh: Is it possible to add this within the main parser instead of adding an extra loop outside here?

Bumsparkle: Done

pyzx/tikz.py
```diff
@@ -220,6 +246,41 @@ def tikzit(g: Union[BaseGraph[VT,ET],Circuit,str], draw_scalar:bool=False) -> No
 synonyms_wedge = ['w edge', 'w io edge']
 
 tikz_error_message = "Not a valid tikz picture. Please use Tikzit to generate correct output."
+
+# Supports both "% zxlive-metadata: <base64payload>" and "% pyzx-metadata: <base64payload>"
+_TIKZ_METADATA_RE = re.compile(
+    r"^\s*%\s*(?:zxlive-metadata|pyzx-metadata):\s*(\S+)\s*$"
```
RazinShaikh: I think zxlive-metadata is not really necessary since the to_tikz method produces pyzx-metadata string only.

Bumsparkle: Agreed

dlyongemallo:
Does this work with H-boxes, and Z-boxes with symbolic phases?

pyzx/tikz.py
```diff
@@ -342,6 +407,18 @@ def handle_phase_error(msg: str) -> None:
...
             set_phase(v,1)
         elif label:
+            phase_dot_styles = {"z phase dot", "x phase dot", "z dot", "x dot"}
+            if (
+                ty in (VertexType.Z, VertexType.X)
+                and style.lower() in phase_dot_styles
+                and any(ch.isalpha() for ch in label)
+            ):
+                try:
+                    phase = string_to_phase(label, g)
```
dlyongemallo: Will this actually work? The Tikzit tikz labels use LaTeX notation, whereas `string_to_phase` expects plain text as a multiple of `pi`. So you'd have to convert, e.g., `\frac{\pi}{4}` to `1/4`.

pyzx/tikz.py
```diff
@@ -342,6 +407,18 @@ def handle_phase_error(msg: str) -> None:
         elif label == r'\neg':
             set_phase(v,1)
         elif label:
+            phase_dot_styles = {"z phase dot", "x phase dot", "z dot", "x dot"}
+            if (
+                ty in (VertexType.Z, VertexType.X)
+                and style.lower() in phase_dot_styles
+                and any(ch.isalpha() for ch in label)
```
dlyongemallo: I don't think this guard is correct. The string `\frac{\pi}{4}` contains the letters `f`, `r`, etc. This causes `string_to_parse` to fail and fall through to the old parsing code (which I think does output the correct result in this case, but by accident). But it might misparse something as a variable name in some cases.

In fact, instead of a guard here, the LaTeX needs to be converted to plain text first.

Bumsparkle: I could remove that guard and instead do deterministic LaTeX-to-plain normalisation first, then parse?

pyzx/tikz.py
Bumsparkle: Good point. I could fix this by normalising LaTeX phase labels to plain phase expressions first (e.g. `\frac{\pi}{4}` -> `1/4`) before calling string_to_phase?

pyzx/tikz.py
```diff
@@ -342,6 +506,23 @@ def handle_phase_error(msg: str) -> None:
         elif label == r'\neg':
             set_phase(v,1)
         elif label:
+            phase_dot_styles = {"z phase dot", "x phase dot", "z dot", "x dot"}
+            normalised_label = _normalise_tikzit_phase_label(label)
+            if (
+                ty in (VertexType.Z, VertexType.X)
```
dlyongemallo: Is this necessary? It doesn't support H-box, and a Z-box with a symolic phase will end up skipping the new parser path.

pyzx/tikz.py
```diff
@@ -342,6 +506,23 @@ def handle_phase_error(msg: str) -> None:
         elif label == r'\neg':
             set_phase(v,1)
         elif label:
+            phase_dot_styles = {"z phase dot", "x phase dot", "z dot", "x dot"}
```
dlyongemallo: Won't this break for non-phase-dot styles? Add a test with `style=z spider` or `style=green dot` (not just `z phase dot`).

In fact, why gate anything here at all? Why not just replace the old parsing code unconditionally with your new code, since you normalise beforehand? I think your tests actually show that the new parsing code handles everything, but it's gated behind a check that prevents it from being used in many scenarios when it should handle them perfectly fine.

pyzx/tikz.py
Bumsparkle: You're right, I've removed it so the normalised `string_to_phase` path is now used generally.

pyzx/tikz.py
Bumsparkle: Yep the style check is unnecessarily blocking valid cases like z spider and green dot, so I removed it and made the normalised parser path the primary one after normalisation. I also added tests for those non-phase-dot styles, plus H-box and Z-box symbolic labels, to show the broader path is actually exercised.

pyzx/tikz.py
```diff
@@ -220,6 +246,140 @@ def tikzit(g: Union[BaseGraph[VT,ET],Circuit,str], draw_scalar:bool=False) -> No
...
+    payload_str = match.group(1)
+    lines_without_metadata = lines[:line_index] + lines[line_index + 1:]
+    clean_tikz = "\n".join(lines_without_metadata)
+    try:
+        payload_bytes = base64.urlsafe_b64decode(payload_str.encode("ascii"))
+        metadata = json.loads(payload_bytes.decode("utf-8"))
+        variable_types = metadata.get("variable_types", {})
+        if isinstance(variable_types, dict):
+            return clean_tikz, {str(k): bool(v) for k, v in variable_types.items()}
+    except Exception:
```
dlyongemallo: Narrow this to `ValueError`. Otherwise this will silently swallow things like `KeyError` or `AttributeError` thrown by `string_to_phase` internals, which we do want to actually show.

pyzx/tikz.py
```diff
@@ -145,10 +149,32 @@ def _to_tikz(g: BaseGraph[VT,ET], draw_scalar:bool = False,
...
+        str(name): bool(g.var_registry.get_type(name, False))
+        for name in g.var_registry.vars()
+    }
+    if not variable_types:
+        return None
+    payload = json.dumps({"variable_types": variable_types}, separators=(",", ":")).encode("utf-8")
+    return base64.urlsafe_b64encode(payload).decode("ascii")
+
+
+def to_tikz(g: BaseGraph[VT,ET], draw_scalar:bool=False, preserve_metadata:bool=True) -> str:
```
dlyongemallo: Is the `preserve_metadata` flag necessary, i.e., will there ever be a reason to not just leave it as `True` under any real scenario?

dlyongemallo (commented):
I would suggest adding a metadata round-trip test: a graph with symbolic phase should survive the `to_tikz` followed by `tikz_to_graph` with variable types preserved.

Also: I think you can remove the old parser code path entirely. It's only called if normalisation returns `None` _and_ the label happens to contain `pi` or is a Z-box, but Z-box with complex labels are already handled before either parser runs. If the label has `pi`, but also unrecognised characters, the old parser fails which is the same outcome in the new parser. Otherwise, any pattern the old parser can successfully parse is already converted by the normaliser into something `string_to_phase` can handle. So AFAICT the old parsing path does nothing after your change, except to make the route to `handle_phase_error` longer for unparseable labels.

But it's possible I missed some corner cases. Please check carefully if the new parser can indeed handle anything the old parser could, and if so, consider removing the latter entirely to keep things simple.

pyzx/tikz.py
RazinShaikh: The only reason I can think of is backward compatibility. For example, someone might save a file to TikZ using a newer version of PyZX and then try to open it with an older version. The older version may not recognize the metadata comment and could fail to parse it, resulting in an error.

Bumsparkle: `preserve_metadata` was for backward compatibility: if a newer PyZX writes metadata as a separate leading comment line, an older PyZX version that expects the first non-empty line to start with `\begin{tikzpicture}` might reject the file entirely. If preserving compatibility is needed I can either keep the flag for that case or I could change the encoding so the metadata is attached as an inline `% ...` comment on the `\begin{tikzpicture}` line, which then older parsers are more likely to tolerate?

dlyongemallo: I don't understand how this helps with backwards compatibility. If you had a tikz file wtih symbolic phases, like `$theta$`, in the labels, an older version of `pyzx` would fail while trying to parse it as a numeric phase. The data format has changed, so whether it breaks on the first line or somewhere else, it's going to break, and arguably it might be better to break the same way on the first line for all such files than to break in a different place depending on the specific file. I think the older version not recognising the metadata comment and failing to parse is _the correct behaviour in this situation_.

It's valid for a `%` to begin a LaTeX file, so it was a bug with older pyzx versions not to tolerate this. Even if you stuck the `%` somewhere else, and an older pyzx parsed past the header, what's it going to do with the actual symbolic phases in the file? 

I think the "fail fast' principle applies here: the user should be informed immediately that they have a file that their version of `pyzx` wasn't designed to handle, instead of potentially silently losing data. So my recommendation is to remove the `preserve_metadata` flag and always add the payload when it exists, and indicate in the commit description that this PR introduces a breaking change to the file format, and also note that in the CHANGELOG.

Bumsparkle: That makes sense. I can remove the `preserve_metadata` flag, always emit the metadata when it exists and note in the commit description and changelog that this is a breaking change to the TikZ round-trip format for symbolic metadata.

pyzx/tikz.py
Bumsparkle: I narrowed the metadata parsing path to ValueError only

PR #438, opened 2026-04-28
Refactor: Remove Galois dependency and replace with Mat2 from linalg.py

### LINKED ISSUE #437: Replace `GF2` with the internal `linalg.py`
The pauli webs computed in https://github.com/zxcalc/pyzx/blob/master/pyzx/web/firing_assignments.py use GF2 from the galois package. But PyZX already has an implementation of linear algebra with F2 matrices in linalg.py. So adding the galois package dependency is unnecessary and it's better to use the internal implementation to avoid bloat. I noticed this issue when trying to compile and build binaries for ZXLive; galois uses numba for just-in-time compilation which is incompatible with ahead-of-time for building binaries

RazinShaikh:
Remove Galois dependency and replace with Mat2 from linalg.py

Fixes https://github.com/zxcalc/pyzx/issues/437

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Hence, occasionally changes will be backwards incompatible (although they will a
 
 ### Removed
 - `Measurement.to_graph_ground`; measurements now always use the symbolic boolean representation (by @dlyongemallo).
+- The `galois` package dependency. PyZX now uses the internal `Mat2` class for computing Pauli webs in `pyzx.web`.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


jvdwetering:
Huh, all the functionality needed was already in Mat2? Very interesting

RazinShaikh:
the only missing bit was the XOR but that was not too difficult to implement manually

maximilianruesch:
The reason I chose `galois` over `Mat2` was for speed of solving the nullspace, which JIT really helps with. But especially for small graphs (as expected to be handled with ZXLive) this is perhaps not a concern.

PR #444, opened 2026-05-09
Enforce type annotations for all functions

96-LB:
This PR is a work-in-progress of an effort to add type annotations to all functions and to enable a mypy setting which enforces type annotations for all future contributions.

This is a rather large undertaking that affects every corner of the codebase. I'm doing my best not to change the actual functionality of any of the code, but adding type annotations often results in large cascades of errors due to improperly typed variables which were being hidden/suppressed before (for example, mismatches between base classes and derived classes, or invalid function calls in legacy code), and it sometimes becomes unavoidable to rewrite a few lines to satisfy the typechecker.

For that reason, I think it's probably important to treat this PR with close scrutiny to make sure no serious breaking changes are accidentally introduced. I've opened it as a draft early so that progress can be monitored, but it's far from complete at the moment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ pyzx = ["js/*.js", "py.typed"]
 python_version = "3.10"
 warn_return_any = false
 warn_unused_configs = true
+disallow_untyped_defs = true
 
 [[tool.mypy.overrides]]
 module = "pyzx.optimize"
--- a/pyzx/altextract.py
+++ b/pyzx/altextract.py
@@ -1,4 +1,3 @@
-from fractions import Fraction
 import itertools
 import math
 
@@ -8,7 +7,7 @@ from .simplify import pivot_simp
 from .circuit import Circuit
 from .graph.base import BaseGraph, VT, ET
 
-from typing import Dict, Union
+from typing import Dict, Sequence, Union
 
 from .extract import bi_adj, connectivity_from_biadj, max_overlap
 # , greedy_reduction, find_minimal_sums, xor_rows, column_optimal_swap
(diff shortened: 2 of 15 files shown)


96-LB:
Some notes from the first round of updates:
- Especially in the `scripts` folder, there was a lot of code which was very outdated or even no longer functional. I started updating some of it but then decided it would be best to hold off on that for now and focus on more important code. 
- Graphs are often difficult to properly type because the output of Graph() is a BaseGraph but a lot of code (even inside of BaseGraph itself!) implicitly relies on a specific backend's particular implementation. It may be worth refactoring the way backends work in the future if more type-safety is desired.

dlyongemallo:
> This is a rather large undertaking that affects every corner of the codebase.

Can you maybe split this into multiple commits, each of which enforces type annotations on smaller sections of code?

96-LB:
Yes, that seems wise. The work I did today was scattered as I just went down the list of issues, but seeing the vast scope of the changes that still need to be made, it would probably be better to move forward more systematically.

Are there any areas that you think should be focused on first?

jvdwetering:
If possible I would try to make separate PRs if you are going, say, file by file. That might also make it easier to deal with merge issues in case it takes a while to get this thing done.

jvdwetering:
You might also want to take #431 into account and see whether removing it in certain files, or replacing it by imports from `typing` instead breaks anything on the lowest supported Python version.

jvdwetering:
And this is just a note for myself: I was a little surprised you were type annoting a subclass of a generic class, but I looked online and this does seem to be the idiomatic way to do it. From what I understand it will then also give an error if the inferred type from subclassing the generic one contradicts the explicitly defined type.

PR #446, opened 2026-05-18
Add Gaussian distribution CNOT spread

mjsutcliffe99:
When generating random circuits via the pyzx.generate functions, every CNOT places its control and target on separate randomly selected qubits (with equal probability weighting to all qubits, i.e. on a uniform distribution). However, it can be useful to generate random circuits whose CNOTs favour short-range connections (i.e. connecting nearby qubits) as there are some arguments to suggest such circuits are more realistic.

As such, switching from effectively a uniform distribution to a normal/Gaussian distribution to determine the span of each CNOT would be handy. The standard deviation, sigma, (such that sigma^2 is the variance) can be included in the circuit generation functions as an extra optional argument (which effectively defaults to infinity, i.e. a normal distribution, so these functions behave exactly as before when sigma is not explicitly specified).

sigma=float("inf") (or sigma=None) means a uniform distribution and sigma=0 means every CNOT will connect to an immediate neighbour qubit, and e.g. sigma=1.5 means high preference for short-range CNOTs. See bottom of page 12 and figure 7a in https://arxiv.org/pdf/2409.00828 for extra clarity.

Examples:
image
image
image
image

### DIFF
--- a/pyzx/generate.py
+++ b/pyzx/generate.py
@@ -27,7 +27,7 @@ __all__ = [
     "qft",
 ]
 
-import random, types
+import random, types, math
 from fractions import Fraction
 
 from typing import Optional, List, Set, Union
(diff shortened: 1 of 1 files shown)


jvdwetering:
I'm guessing this is part of you porting all the functionality from paramZX?
Looks good to me.

PR #451, opened 2026-05-25
Fix merge_phase_gadget_rule for boolean Poly phases

### LINKED ISSUE #436: Bug in rewrites with boolean parameters
@Hope-Alemayehu and I were looking into these kinds of issues anyways so we're happy to try this. In any case, here's documenting the issue to have a record of what we're fixing and why.

## The problem
PyZX doesn't support rewrites for parametrized boolean rewrites. It does produce output, but the output is incorrect.

### A test case that fails
The QASM file for T injection below, which parses to the below circuit:
```
circ = zx.qasm("""
OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg c[1];
reset q[1];
h q[1];
t q[1];
cx q[0],q[1];
measure q[1] -> c[0];
if (c==1) s q[0];
""")
zx.draw(circ)
```

Image

When you then call
```
g = circ.to_graph()
zx.full_reduce(g)
zx.draw(g)
```
the output is

Image
which is clearly incorrect. The correct phase is (-1)^(c[0]) * pi/4 + (1/2) * c[0] * pi, which equals pi/4 regardless of whether c[0] = 0 or 1.

### The partial solution
The most likely guess as to the issues, is that the pi phase commute rule (between red c[0] pi and green pi/4) here is not prefixed by (-1)^(c[0]).

## The full but more complicated solution
Each of the below rules should be implemented and verified. This might need coming up with less janky ways to deal with symbolic and global phases.
Image

Hope-Alemayehu:
Fixes #436.

The `match_phase_gadgets` function used `phases[n] != 0` to check whether an axel's phase is pi before negating the gadget phase. For `Fraction` phases this works correctly, but for `Poly` phases the check tests whether the polynomial object is non-empty, which is always true regardless of what the boolean variable evaluates to. This caused unconditional negation even when `c[0]=0` should have left the phase unchanged.

The fix replaces the unconditional `-phases[v]` with:

```python
gp + (-gp - gp) * phases[n]
```

This evaluates to `gp` when `phases[n]=0` and `-gp` when `phases[n]=1`, keeping the boolean variable alive without introducing a coefficient of 2. The same pattern is applied to the `totphase` computation in the multi-gadget case. A test using the T injection circuit from the issue is also included.

Happy to make any changes if needed.

cc @lia-approves

### DIFF
--- a/pyzx/rewrite_rules/merge_phase_gadget_rule.py
+++ b/pyzx/rewrite_rules/merge_phase_gadget_rule.py
@@ -26,7 +26,7 @@ from typing import Optional
 
 from pyzx.utils import  FractionLike, phase_is_pauli
 from pyzx.graph.base import BaseGraph, VT, ET
-from pyzx.symbolic import Poly
+from pyzx.symbolic import Poly, new_const
 
 
 __all__ = ['merge_phase_gadgets_for_simp',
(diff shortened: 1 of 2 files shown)


pyzx/rewrite_rules/merge_phase_gadget_rule.py
```diff
@@ -85,12 +86,30 @@ def match_phase_gadgets(g: BaseGraph[VT,ET], vertices:Optional[List[VT]]=None) -
         if len(gad) == 1:
             n = gad[0]
             v = gadgets[n]
-            if phases[n] != 0: # If the phase of the axel vertex is pi, we change the phase of the gadget
+            if phases[n] != 0 or isinstance(phases[n], Poly): # If the phase of the axel vertex is pi, we change the phase of the gadget
```
dlyongemallo: I think this check might be too broad, as it matches any `Poly` including a constant one with no free variables. Maybe check `len(phases[n].free_vars() > 0)` for the symbolic branch (and leave constants alone in the original branch).

Hope-Alemayehu: oh okay, I've made the changes.

pyzx/rewrite_rules/merge_phase_gadget_rule.py
```diff
@@ -85,12 +86,34 @@ def match_phase_gadgets(g: BaseGraph[VT,ET], vertices:Optional[List[VT]]=None) -
...
+            if axel_phase != 0 or is_symbolic:
                 g.scalar.add_phase(phases[v])
                 g.phase_negate(v)
-                m.append((v,n,-phases[v],[],[]))
+                if is_symbolic:
+                    if isinstance(phases[v], Poly):
+                        gadget_phase: FractionLike = phases[v]
+                    else:
+                        gadget_phase = new_const(cast(Union[int, Fraction], phases[v]))
+                    neg_gadget = -gadget_phase
```
jvdwetering: neg_gadget is only used in the next line, why not say `-2*gadget_phases` there instead?

Hope-Alemayehu: yes, it's better to write it like that. I've updated it.

jvdwetering:
Thanks for the fix!

Hope-Alemayehu:
> Thanks for the fix!

thanks for the review:)

Issue #457, opened 2026-05-29
``full_reduce`` not working properly with non-Clifford spiders


MSzynisz:
When I run this simple example, I get an error:
```
import pyzx as zx
c = zx.Circuit(1)
c.add_gate("ZPhase",0,phase=0.124312)
g = c.to_graph()
zx.full_reduce(g)
```
Here is the error:
```
Traceback (most recent call last):
  File "/home/mszynisz/try.py", line 5, in <module>
    zx.full_reduce(g)
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/simplify.py", line 218, in full_reduce
    clifford_simp(g)
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/simplify.py", line 179, in clifford_simp
    i2 = pivot_boundary_simp(g)
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/rewrite.py", line 44, in __call__
    return self.simp(graph)
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/rewrite.py", line 256, in simp
    return  self.simp_applier(graph)
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/rewrite_rules/pivot_rule.py", line 107, in pivot_boundary_for_simp
    matches = match_pivot_boundary(g)
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/rewrite_rules/pivot_rule.py", line 153, in match_pivot_boundary
    if types[v] != VertexType.Z or not phase_is_pauli(phases[v]) or g.is_ground(v):
  File "/home/mszynisz/.local/lib/python3.10/site-packages/pyzx/utils.py", line 143, in phase_is_pauli
    return phase.is_pauli
AttributeError: 'float' object has no attribute 'is_pauli'
```
The issue seems to be the float phase. This works in v.0.9.0, but does not work in the current version, and ``phase_is_pauli`` seems not to have changed between the two versions.

dlyongemallo:
The traceback shows where the error was introduced: [`pivot_rule.py` line 153](https://github.com/zxcalc/pyzx/blob/ad802d41ebbbfe627911adde8e728f2d818a729c/pyzx/rewrite_rules/pivot_rule.py#L153). This line is currently:

```python
if types[v] != VertexType.Z or not phase_is_pauli(phases[v]) or g.is_ground(v):
```

[...]

PR #459, opened 2026-05-31
add `settings.strict_phase_types` to reject float phases at graph entry

dlyongemallo:
## Description

The library implicitly accepts float phases, which only worked by accident in many code paths. `full_reduce` crashed with `AttributeError` on a spider carrying a float phase because rewrite rules assumed `FractionLike`. 

Floats reaching `set_phase`/`add_to_phase` are now rejected with a `TypeError` by default, surfacing a clear error at the entry point instead of crashing deep inside a rewrite rule. 

Setting `pyzx.settings.strict_phase_types = False` opts in to automatic conversion via `settings.float_to_fraction_max_denominator`.

## Related issue

#457

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Hence, occasionally changes will be backwards incompatible (although they will a
 
 ### Added
 - Added the "magic cat" decomposition strategy from https://arxiv.org/pdf/2202.09202 (which previously only existed in Quizx). (by @mjsutcliffe99).
+- `settings.strict_phase_types` (default `True`) rejects float phases at `set_phase`/`add_to_phase` rather than letting them flow into the graph and crash downstream rewrite rules. Set to `False` to opt in to automatic conversion to `Fraction` using `settings.float_to_fraction_max_denominator`, accepting the resulting precision loss. Fixes crash in `full_reduce` with non-Clifford spiders (by @dlyongemallo).
 
 ## [0.10.3] - 2026-06-01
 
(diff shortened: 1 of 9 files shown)


jvdwetering:
So one problematic thing is that if someone gives a chemistry circuits with very small floating point angles, then this will silently round those phases to nearest fraction. Maybe there should a setting, set by default to True, of being 'strict' so that normalize_phase would give an exception when given a float, with an error message explaining that strict has to be set to False and that this rounds the phase to nearest fraction based on the other setting.

dlyongemallo:
Under strict mode, what should happen to float phases which are exactly Clifford, e.g. `setphase(v, 0.5)`?

dlyongemallo:
I've updated the PR with a `strict_phase_types` setting as suggested.

jvdwetering:
> Under strict mode, what should happen to float phases which are exactly Clifford, e.g. `setphase(v, 0.5)`?

That's a good question. If there is some easy way to check whether a floating point is an exact dyadic rational then you can just convert it to a Fraction without any precision loss. I guess all floats are implicitly a dyadic rational though, so maybe it is just that if it is a dyadic rational with a particularly small numerator that it should just accept. Maybe something like `phase == float(Fraction(phase))`?

dlyongemallo:
I thought about this as well as experimented a bit, and concluded that it would lead to surprises/weird edge cases/bugs if the `strict_phase_types` rule had exceptions, e.g., `0.5000....001` behaves differently depending on how it's rounded and what operations are performed. The most consistent (and least bug-prone) application of the rule would be to take it literally. So `0.5` is a `float` and should be rejected when the strict flag is set, and if the user actually wants `Fraction(1,2)`, they can explicitly write that instead.

PR #461, opened 2026-05-31
Improve type annotations in `pyzx.graph` submodule

96-LB:
Smaller-scoped version of #444 targeting just the graph classes.

This PR is part of an effort to introduce type annotations on all functions across the library and enable the mypy setting to enforce it on all future contributions. I also updated some annotations to slightly more modern syntax since the lowest supported python version has now been raised to 3.10.

For the most part, I avoided touching the bodies of functions, but there's a few areas where I slightly modified things in order to fix type errors (e.g., adding `list(...)` at call sites, or removing hardcoded references to the `Multigraph` implementation inside `BaseGraph`). There are still some areas where code that should be generic instead relies heavily on specific implementations in a non-typesafe manner, and the currently recommended `Graph()` function is, but fixing these larger issues seems to require a larger refactor which is outside the scope of this PR.

I did not touch `graph_gt.py` or `graph_ig.py` since they are marked as deprecated, but if it's worth also typing these then I can do that in this PR too.

There is one unresolved issue I would like to seek feedback on: currently, `BaseGraph` defines a `edge_set` function which returns a `set[ET]`, but `Multigraph` overrides this to return a `Counter[ET]` instead, which is not a compatible type at all. The solutions to this are either to change what `Multigraph.edge_set` returns (which would be a breaking change) or make `BaseGraph.edge_set` have a uselessly generic type annotations. I didn't want to presume which solution would be better here.

### DIFF
--- a/pyzx/altextract.py
+++ b/pyzx/altextract.py
@@ -94,8 +94,9 @@ def alt_extract_circuit(
     qs = g.qubits() # We are assuming that these are objects that update...
     rs = g.rows()   # ...to reflect changes to the graph, so that when...
     # ty = g.types()  # ... g.set_row/g.set_qubit is called, these things update directly to reflect that
+    qubit_count: int = math.ceil(g.qubit_count()) # We take the ceiling because legacy code allows qubit_count to be a float
     phases = g.phases()
-    c = Circuit(g.qubit_count())
+    c = Circuit(qubit_count)
 
     gadgets = {}
     inputs = g.inputs()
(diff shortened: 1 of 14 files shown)


jvdwetering:
For the edge_set function there would be multiple ways to solve this: make edge_set always return a set, so that its behaviour changes for multigraph, and then add a new function which recovers the functionality for multigraph, which would be identical to edge_set for simple graphs. However, this has the issue that all the code has to be updated to use this new function.

The other solution is to indeed give it a quite useless broad type in the BaseGraph implementation and specify this more specifically in GraphS and MultiGraph. I think for now this is the best solution, since the goal of PRs like this is just to annotate the current behaviour.

96-LB:
Okay, I changed the type annotation to `set[ET] | Counter[ET]` on the base type. The areas where the function were used seemed to all eventually convert to Counter anyway, so I tweaked the callsites to fix the resulting mypy errors.

This PR is probably ready for review/merge at this point, unless anybody feels that `graph_gt.py` and `graph_ig.py` should be given type annotations now. In general I plan to de-prioritize deprecated code or legacy scripts as I go through the codebase, so that the important stuff can be handled first.

jvdwetering:
Thanks!

Issue #413, opened 2026-03-02
Boundary pivot and gadget pivot do not match


RazinShaikh:
Boundary and gadget pivot are defined as equations (P2) and (P3) on page 6 of https://doi.org/10.1103/PhysRevA.102.022406

Recreating the same diagrams in ZXLive do not match for these rule. For example, here's a diagram from boundary pivot:

<img width="2300" height="1708" alt="Image" src="https://github.com/user-attachments/assets/d601d1e2-f89d-49cb-a4d8-c391cf4561af" />

dlyongemallo:
> equations (P2) and (P3) on page 6

Can I assume these are identical to (P2) and (P3) on p.13 of the [arXiv preprint](https://arxiv.org/abs/1903.10477)? I don't have access to the APS version.

The API for the current implementations of `pivot_boundary_simp` and `pivot_gadget_simp` seem inconsistent with `pivot_simp`. The latter is a `RewriteSimpDoubleVertex` and the former two are `RewriteSimpGraph` (see [here](https://github.com/zxcalc/pyzx/blob/b447545bbfc4269bed9fac0e1b652c1a9e2a12ee/pyzx/simplify.py#L69-L76)). I think they should be rewritten as `RewriteSimpDoubleVertex`, since that already supports both `simp()` on the whole graph and `apply()` on each pair. That would make the logic cleaner and I think easier to understand, but it would be a breaking API change.

RazinShaikh:
yes, it's identical to the arxiv version

PR #462, opened 2026-06-01
fix bug in `match_pivot_boundary` which skipped exclusion of grounded neighbours

dlyongemallo:
## Description

`match_pivot_boundary` had a typo in the check for grounded neighbours (`consumed_vertices` evidently miscopied from the previous line). This fixes the check.

## Related issue

Fixes a bug that partly contributes to #413.

## Checklist


dlyongemallo:
Split off from the larger change #460, as this is a clear bug in isolation.

PR #473, opened 2026-06-30
Improve type annotations and fix type errors in `pyzx.circuit` submodule

96-LB:
Adds missing type annotations to the files in `pyzx/circuit/` and fixes a few related problems. Also modernises some of the syntax for Python 3.10 (e.g. using `list` and `dict` instead of `typing.List` and `typing.Dict`) and removes all instances of `#type: ignore`. Function bodies are mostly untouched, with some exceptions to fix type errors:
- A lot of the code in the base `Gate` class used `hasattr` checks and then accessed properties directly, which the type checker doesn't like. This PR replaces these with explicit `getattr` calls so that it's clear we're circumventing the usual behaviour. This is not any safer at runtime, but it at least signals intent and avoids mypy errors (which were previously ignored with broad `#type: ignore`s which could also hide other errors). This code should really be refactored so that subclasses override behaviour instead of relying on `hasattr` in the base class, but this is outside the scope of this PR.
- In some subclass overrides of `Gate.to_basic_gates()`, integer phases could incorrectly become floats because they were being divided by 2. A new utility function `half_phase` is introduced in `pyzx/utils.py` which handles this type error by casting to `Fraction`.
- Some local imports are moved to be global because they were (incorrectly) used in type hints. Doing so does not introduce any cyclic dependencies.
- Rather than identifying gates by their `Gate.name` property, which provides no information to the type-checker, `isinstance` checks are performed. This is also more robust, since the names could have collisions or even be changed by code.

This last bullet point requires #472 to be merged first, as the outdated version of `mypy` currently being used doesn't properly perform the needed type narrowing, so this PR is marked as a draft for now.

### DIFF
--- a/pyzx/circuit/__init__.py
+++ b/pyzx/circuit/__init__.py
@@ -15,13 +15,12 @@
 # limitations under the License.
 
 import os
-from typing import List, Union, Optional, Iterator, Dict
+from typing import Union, Iterator
 
 import numpy as np
 
-from .gates import (Gate, gate_types, NOT, Y, Z, HAD, XPhase, YPhase, ZPhase, U2, U3, S, T, SX, SWAP, RXX, RZZ, CNOT,
-                    CY, CZ, CHAD, CSX, XCX, CRX, CRY, CRZ, CPhase, CU3, CU, CSWAP, Tofolli, CCZ, ParityPhase, FSim,
-                    Measurement, PhaseGadget, ConditionalGate)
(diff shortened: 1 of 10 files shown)


96-LB:
Since the #472 was merged, I've marked this PR as ready. There were some merge conflicts with other PRs in the QASM parser, but I fixed those and updated the new type annotations to the modern format. I also updated the `__all__` in `__init__.py` to reflect the types which are being exported but are otherwise unused in that file (because I mistakenly removed some unused imports and broke some tests).

jvdwetering:
Very nice, thanks!

PR #475, opened 2026-07-01
Fix numeric phases becoming a constant `Poly` after a TikZ round trip

### LINKED ISSUE #471: Fractional phases silently become a constant `Poly` after a `to_tikz`/`tikz_to_graph` round trip
Hi! I've been exploring PyZX, specifically the TikZ <-> PyZX workflow, and came across something that seemed off.

If I export a diagram with a fractional phase (e.g. π/4) to TikZ and import it back, the phase comesback as a constant `Poly` with no free variables, instead of a `Fraction`. The value is still correct (`== Fraction(1, 4)` returns `True`) and nothing errors at import, so it looks fine at first

It surfaces later, when computing on the diagram as `to_tensor()` raises a `ValueError` ("Can't convert diagram with parameters to tensor: 1/4"). `full_reduce` and `extract_circuit` don't seem to complain, but a circuit extracted from such a diagram seems to be silently affected as its own `to_tensor()` raises the error. For contrast, a JSON round trip of the same diagram keeps the phase as a `Fraction`

Minimal repro:

```python
import pyzx as zx
from pyzx import tikz
from fractions import Fraction

print("pyzx version          :", zx.__version__)

c = zx.Circuit(1)
c.add_gate("ZPhase", 0, phase=Fraction(1, 4))
g = c.to_graph()
print("original phase type   :", type(g.phase(1)).__name__)

g2 = tikz.tikz_to_graph(tikz.to_tikz(g))
g2.auto_detect_io()
print("after TikZ round-trip :", type(g2.phase(1)).__name__)
print("value still correct   :", g2.phase(1) == Fraction(1, 4))

g3 = zx.Graph.from_json(g.to_json())     # JSON round-trip, for contrast
print("after JSON round-trip :", type(g3.phase(1)).__name__)

try:
    g2.to_tensor()
except ValueError as e:
    print("to_tensor() fails     :", e)
```

Output (current master):

```
pyzx version          : 0.10.3
original phase type   : Fraction
after TikZ round-trip : Poly
value still correct   : True
after JSON round-trip : Fraction
to_tensor() fails     : Can't convert diagram with parameters to tensor: 1/4
```

It looks like `tikz_to_graph` routes the phase label through `string_to_phase`, and for a numeric label that ends up as a constant `Poly` rather than collapsing back to a `Fraction` (seemingly the same fallback that handles symbolic phases).

Is this expected, or should a phase with no free variables come back as a `Fraction` after a tikz round trip? If it's worth changing, I have a couple of fixes in mind, like adjusting the TikZ phase normaliser, or collapsing a parameter free `Poly` to a `Fraction` in `string_to_phase`. I'd be happy to open a PR with a regression test if that would help.

Thanks!

gauthamkanagaraj:
## Description

A numeric spider phase (e.g. π/4) written out with `to_tikz` and read back with `tikz_to_graph` came back as a constant `Poly` instead of a `Fraction`. The value was correct and compared/printed as expected, so it was easy to miss, but a parameter-free `Poly` still counts as having parameters, so the round tripped graph raised `Can't convert diagram with parameters to tensor` when passed to `to_tensor`/`to_matrix`.

The TikZ phase normaliser emits numeric fractions as `(num)*1/den`, a form `string_to_phase` only parses through its symbolic fallback (yielding a `Poly`). Rather than special-casing this in the TikZ importer, the fix moves upstream of every caller. After parsing, `string_to_phase` now collapses a variable free parse result to a numeric value using the existing `Scalar.simplify_poly` helper, returning a Fraction for a numeric result and leaving genuinely symbolic phases as a Poly. Symbolic phases are unchanged and still produce a `Poly`.

Regression tests round trip a numeric phase and assert a `Fraction` result, including through the phase-keyed dicts in `Scalar.from_json` (equal phases must hash equally as dict keys). The existing symbolic and suffixed round-trip tests are retained.

## Related issue

Fixes #471

## Checklist
- [x] For new features: I have updated the documentation. (N/A - bugfix, no new feature)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Hence, occasionally changes will be backwards incompatible (although they will a
 - Added the "magic cat" decomposition strategy from https://arxiv.org/pdf/2202.09202 (which previously only existed in Quizx). (by @mjsutcliffe99).
 - `settings.strict_phase_types` (default `True`) rejects float phases at `set_phase`/`add_to_phase` rather than letting them flow into the graph and crash downstream rewrite rules. Set to `False` to opt in to automatic conversion to `Fraction` using `settings.float_to_fraction_max_denominator`, accepting the resulting precision loss. Fixes crash in `full_reduce` with non-Clifford spiders (by @dlyongemallo).
 
+### Fixed
+- A numeric phase (e.g. `π/4`) now survives a `to_tikz`/`tikz_to_graph` round trip as a `Fraction` instead of becoming a constant `Poly`. The TikZ phase normaliser emitted numeric fractions in a form that `string_to_phase` could only read through its symbolic fallback, producing a parameter-free `Poly` that compared and printed correctly but raised `Can't convert diagram with parameters to tensor` when passed to `to_tensor`/`to_matrix` (by @gauthamkanagaraj).
+
 ## [0.10.3] - 2026-06-01
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,14 +8,14 @@ Hence, occasionally changes will be backwards incompatible (although they will a
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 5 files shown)


jvdwetering:
Thanks for the fix! Can you update the CHANGELOG line so it belongs to the unreleased version?

gauthamkanagaraj:
Hi @jvdwetering, I've updated the CHANGELOG entry now and moved it to `## [Unreleased]`. Thanks!

PR #479, opened 2026-07-07
add support for custom gates with phases in QASM parser

dlyongemallo:
## Description

Accept custom gate definitions with parameters in `Circuit.from_qasm`, which previously raised `TypeError`.

Phase expressions may use sums and products of parameters with rational or `pi`-based constants (e.g., `2*theta`, `theta + pi/4`). 

Division (`theta/2`) and bracketed expressions (`(theta+pi/2)/2`) aren't supported (yet).

## Related issue

#220, #469, #478

## Checklist
- [x] I have added/updated tests (for bugfixes, please include a regression test, for new features, include new tests either to an existing test file or a new file).
- [x] For new features: I have updated the documentation.
- [x] All the functions I added have a docstring parsable by sphinx (for a good example see `pyzx.extract.extract_circuit`).
- [x] I have added an entry to CHANGELOG.md describing my change.

Issue #220, opened 2024-05-04
support custom gates with phases


dlyongemallo:
In particular, I'd like to use Qiskit's `random_circuit` function, but it generates custom gates with phase parameters. For example:

```python
from pyzx.circuit import Circuit
from qiskit.circuit.random import random_circuit
from qiskit.qasm2 import dumps
...
qiskit_circ = random_circuit(qubits, depth)
qasm = dumps(qiskit_circ)
circ = Circuit.from_qasm(qasm)
```

This generates code such as:

```openqasm
OPENQASM 2.0;
include "qelib1.inc";
gate r(param0,param1) q0 {
    u3(param0,param1 - pi/2,pi/2 - 1.0*param1) q0;
}
qreg q[1];
r(pi/2,pi/4) q[0];
```

dlyongemallo:
The above raises a `TypeError` in [this line](https://github.com/Quantomatic/pyzx/blob/107c24e1a9869da6dde20404a73e25aefb2b46c1/pyzx/circuit/qasmparser.py#L92-L93).

Geremia:
@dlyongemallo I get the same issue (#478)  when trying to use custom-defined multi-controlled X gates (MCX):

> ```python
> from qiskit import QuantumCircuit, qasm2, qasm3
> 
> qc = QuantumCircuit(6)
> qc.mcx([0, 1, 2, 3, 4], [5]) # Add MCX with 5 control qubits
> qc.draw(output="mpl", filename='cccccx_gate.png')
> ```
> 
>     
>   
> 
[...]

PR #480, opened 2026-07-08
add support for division in symbolic expression grammar

dlyongemallo:
## Description

Change `pyzx.symbolic.parse` to accept division, e.g., `theta/2`, `2*x/4`, `(x + y)/2`, and `x^2/2`, which binds at the same precedence as multiplication. Also fixed operator precedence so that `^` is higher than `*` and `/`.

`Poly.__truediv__` divides integer operands exactly, so dividing by an integer yields a `Fraction` coefficient. 
    
This change lets parametrised custom gate bodies use expressions like `rz(theta/2)`.

Follow-up to #479.

## Related issue

#220

## Checklist

### DIFF
--- a/pyzx/symbolic.py
+++ b/pyzx/symbolic.py
@@ -23,8 +23,7 @@ Lark is used to define a parser that can translate a string into a Poly.
 
 from typing import Any, Callable, Mapping, Union, Optional, Dict, List, Tuple, Set
 from lark import Lark, Transformer
-from functools import reduce
-from operator import add, mul
+from lark.exceptions import VisitError
 from fractions import Fraction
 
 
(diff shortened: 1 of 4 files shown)


jvdwetering:
Just to check: you can only divide by an integer right? Would it make sense to add some tests that make clear that dividing by anything else is not allowed?

pyzx/symbolic.py
```diff
@@ -252,7 +250,11 @@ def __truediv__(self, other: Union['Poly', Fraction, int, float, complex]) -> 'P
         while len(self.terms) > 0 and self.degree >= other.degree:
             leading_term_dividend = sorted(self.terms)[0][1]
             leading_term_divisor = sorted(other.terms)[0][1]
-            coeff = sorted(self.terms)[0][0] / sorted(other.terms)[0][0]
+            num = sorted(self.terms)[0][0]
+            den = sorted(other.terms)[0][0]
+            # Divide exactly for integer operands so that, e.g., `theta/2`
+            # yields a Fraction coefficient rather than a float.
+            coeff = Fraction(num, den) if isinstance(num, int) and isinstance(den, int) else num / den
```
dlyongemallo, line 261: @jvdwetering If the denominator is an integer, the resulting `coeff` is a fraction `Fraction(num, den)`; otherwise, it has value `num / den`. In fact, a `den` like `2.5` (with `num` being `1`) would result in a `coeff` of `Fraction(2,5)` due to `limit_denominator`, and a `complex` value for `den` is also possible.

dlyongemallo: Actually, there are some subtle bugs here when the `den` is a `Fraction` or if it contains `pi`.

dlyongemallo: Oh, and any constant that evaluates to `0`...

dlyongemallo:
@jvdwetering While testing this, I found an existing bug: an expression such as `3/2^2` was evaluated as `9/4` previously, but with my fixes (not yet pushed), it evaluates to `3/4`. Arguably, `3/4` is actually correct (`^` should have higher precedence than `/`), so the change that I'm working on is a bug fix. 

However, it is also a divergence from existing behaviour. Is anyone depending on this behaviour?

jvdwetering:
I'm pretty sure no one is using "^" in these expressions, so feel free to fix the bug.

pyzx/symbolic.py
dlyongemallo: Fixed the division by 0 and chained divsion issues and added tests.

PR #486, opened 2026-07-12
add an `apply_to_boolean_axels` parameter to `unsafe_pauli_push`, defaults to skip boolean Pauli axels

dlyongemallo:
## Description

Pushing a boolean axel through a non-Pauli spider converted a boolean parameter into a non-boolean phase, which might be undesirable behaviour. This adds an opt-in flag `apply_to_boolean_axels` so the user has a choice.

## Related issue

https://github.com/zxcalc/pyzx/issues/436#issuecomment-4338858426

## Checklist

### DIFF
--- a/pyzx/utils.py
+++ b/pyzx/utils.py
@@ -29,6 +29,17 @@ from .symbolic import Poly
 FloatInt = Union[float,int]
 FractionLike = Union[Fraction,int,Poly]
 
+def push_pauli_axel(axel: FractionLike, leaf: FractionLike) -> FractionLike:
+    """Return the new ``leaf`` phase acquired when a Pauli ``axel`` is pushed
+    through a spider of the opposite type. Used by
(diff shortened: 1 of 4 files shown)


pyzx/utils.py
```diff
@@ -29,6 +29,15 @@
 FloatInt = Union[float,int]
 FractionLike = Union[Fraction,int,Poly]
 
+def push_pauli_axel(axel: FractionLike, leaf: FractionLike) -> FractionLike:
+    """Return the new leaf phase after pushing a Pauli ``axel`` through it.
```
jvdwetering: This function looks pretty mysterious on its own. Maybe add a backreference to where this function is used, and that it calculates the phase you get when you push a Pauli through a phase of the opposite type.

dlyongemallo: Done. It'll be less mysterious when other functions call it as part of fixing the rest of #436.

PR #489, opened 2026-07-25
Improve type safety in `pyzx.routing`

96-LB:
Adds type annotations to all functions in the `pyzx.routing` submodule and enables the mypy setting to enforce type annotations in the submodule. Additionall, removes all instances of `#type: ignore` by replacing them with properly type-checked code (with one exception, see below), and modernises the typing syntax.

As in the previous typehint PRs, I did my best to avoid modifying function bodies. However, because of the complex nature of this module, there were several locations where types were inconsistent (including some that seemed like logic bugs) and some code needed updating in order to give type annotations without introducing a plethora of errors. Here's a long list of various comments about the decisions made while annotating the module:

- `Architecture.to_quil_device` seems to be completely broken legacy code—the function body imports `NxDeviceA` from `pyquil.device` and then references `NxDevice`, which seems like a typo. The class `NxDevice` and the module `pyquil.device` have not existed since pyquil v2, which is over 5 years out of date (we are now on v4!). For now, I marked it with a `#type: ignore`, but this function should probably be modernised or completely removed.

- A lot of functions in the module use `**kwargs: Any` to absorb arguments, but this often hides errors, so I would recommend redesigning that. I opted not to change any of those in this PR, but there are many places where the type annotations are a lot less helpful than they could be or type safety is lost because of this. At times (e.g. `Architecture.__init__`), the `**kwargs` parameter is completely unused, and at other times (e.g. `CNOT_tracker.__init__`) the extra arguments are passed into another function, removing all the type safety of the arguments.

- I refactored `FitnessFunction._make_function` to make it easier to type. There's technically a tiny edge case where behaviour could change if `self.metric` is changed between calls to the outputted function (before the function would stay locked to old metric), but this doesn't seem like something that should happen anyway.

- There are various spots in `machine_learning.py` and `phase_poly.py` where null checks were introduced or types were coerced to either lists or numpy arrays, as types were haphazardly mix and matched, or `None` was returned but not handled. In general, all the code that interfaces with numpy was difficult to properly type. Generally, I tried to make the outputs of public-facing functions python lists, and I worried a little less about any internal functions as long as the mappings were consistent.

- `PhasePoly.partition` was an extremely difficult function to decipher because of the way it mixes and matches parities and partitions of parities! I had to refactor the logic within it and its helper functions a little bit in order to separate the two in a way that appeased the type checker, and in doing so I also noticed a couple logic errors (e.g. using `pop` instead of `remove` to remove by value from a list). Because of the high complexity, this is probably the most likely spot that I made an error.

- I replaced references to `Graph` with `GraphS` because the code relied on implementation details. In general, using the `Graph` helper function tends to hurt type safety and opens the codebase up to logic errors which can't be caught statically (e.g. typos in the backend input). I would recommend moving away from that design decision, especially since a lot of the code which instantiates graphs is making assumptions about the implementation anyway.

- A couple of the heuristic functions in `phase_poly.py` were minorly refactored to use for loops instead of iterators, since they were written in an unfriendly way for the type checker (and for human readers).

There were definitely a lot more opinionated changes in this PR than in the previous type-safety PRs, so please feel free to scrutinise the decisions and ask me to change things! In particular, I don't know how this module is used in practice, so if any of the changes I made break things then I can try to solve the typing issues in a different way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,10 @@ module = "pyzx.optimize"
 ignore_errors = true
 
 [[tool.mypy.overrides]]
+module = "pyzx.routing.*"
+disallow_untyped_defs = true
+
+[[tool.mypy.overrides]]
 module = [
     "numpy.*",
     "igraph.*",
--- a/pyzx/generate.py
+++ b/pyzx/generate.py
@@ -1,4 +1,4 @@
-# PyZX - Python library for quantum circuit rewriting 
+# PyZX - Python library for quantum circuit rewriting
 #        and optimization using the ZX-calculus
 # Copyright (C) 2018 - Aleks Kissinger and John van de Wetering
 
@@ -37,6 +37,7 @@ import numpy as np
 from pyzx.circuit.gates import CNOT, ZPhase
 from pyzx.linalg import Mat2, MatLike
 
+from pyzx.routing.cnot_mapper import ElimMode
 from pyzx.routing.parity_maps import CNOT_tracker, Parity
 from pyzx.routing.phase_poly import PhasePoly, mat22partition
 
@@ -76,7 +77,7 @@ def spider(
     outputs: int,
     phase:Optional[Union[FractionLike, complex]]=None,
     ) -> BaseGraph:
-    """Returns a Graph containing a single spider of the specified type 
+    """Returns a Graph containing a single spider of the specified type
     and with the specified number of inputs and outputs."""
     if typ == "Z": typ = VertexType.Z
     elif typ == "X": typ = VertexType.X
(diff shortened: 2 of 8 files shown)


jvdwetering:
I'm just mentioning @Aerylia here in case she would like to take a look at this as well, since it was originally her code. @Aerylia in case you don't want anything to do with this, feel free to let me know :)

jvdwetering:
And let's see if my new Claude setup works

jvdwetering:
@claude review this

96-LB:
@jvdwetering, I've applied most of the suggestions from Claude. It did raise a few concerns about pre-existing behaviour that may be bugged, but I think fixing that is out of scope of this PR.

What are your thoughts on just removing `Architecture.to_quil_device()`? I think it's been completely useless for years, since it does nothing but throw an exception.

jvdwetering:
Yes, I'm fine with removing stuff that obviously doesn't work anymore. But maybe we can do that in a future PR, just to separate it out from this really big PR.

PR #488, opened 2026-07-24
Issue#88 zooming

doczenwiry:
## Description
Added support for zooming in/out of D3 drawings (Jupyter & web browsers). Mouse wheel scroll up/down controls zoom in/out. Clicking the mouse wheel resets the zoom level to the correctly zoomed out version (no cropping of any part that would stick outside the viewing canvas).

## Related issue
This PR is 1/3 of Issue #88.

## Checklist
- I have a test notebook (scratchpads/issue#88.ipynb) with four randomly generated circuits (16 qubits with depth 48, 64 qubits with depth 512, Hadamard edges, non-zero phases & T-gates).
- This feature was tested in Jupyter (running in PyCharm Pro), JupyterLab (running in Safari, Firefox, Chrome).
- The CHANGELOG.md and documentation have been updated.
- I have added some relevant comments in zx_viewer.inline.js for reference.

Issue #88, opened 2022-05-12
Additional convenience features for the d3 drawing of diagrams


jvdwetering:
I suggest two new features for the d3 drawing of diagrams (using zx.draw()):

* When two spiders significantly overlap, there should be some kind of visual indication that allows you to easily see this. For instance, a number in a bright color indicating the number of spiders overlapping, and/or some kind of warning symbol.
* The ability to zoom in and out on a diagram, either by ctrl-scrolling, or by having +/- buttons displayed on the d3 window, whichever turns out to work best.
* If this is possible in Jupyter, the ability to make the display window smaller or bigger in order to show a larger or smaller part of the diagram.

doczenwiry:
Hello everyone,

I'm looking into this issue and I already have a working implementation for the second item using d3-zoom. The wheel scroll-up/down controls the zoom level. I've added a reset of the zoom level when pressing the scroll wheel.

When it comes to the first item, d3 offers a lot of options for force directed layout that would solve the overlap problem without the user having to manually move the spiders around. Would the computational cost be too high for the use-cases envisioned ?

Could you clarify what the third item means; is it that the canvas in which the graph is drawn should be resizable but the zoom level should not change when it is resized ?

jvdwetering:
That's great that you have a working solution for zooming! If you want you can make a separate PR for that.

I don't want to force the 'force directed layout' on users as it might change some carefully manually moved things, and without it I still want users to be aware that things might be overlapping.

About the display window size, currently I believe this is just hardcoded, but it gets annoying when you change the height of a diagram and then it doesn't fit into the current display size. I would consider this a low priority item to fix, the zooming is more important.

doczenwiry:
Thank you ! I'm already working on the collision detection part but I can split out the zooming portion, whichever is lighter on your workflow.

Regarding the tests, I'm using a simple notebook with some randomly generated circuits (16 qubits with depth 48 and 64 qubits with depth 512). Can I put that notebook in the tests/ directory (I only see *.py files in there) or is it better I leave it in the scratchpads/ directory ?

I understand the issue with altering the rest of the layout if it is manually set. D3 seems to offer the possibility of selecting which nodes undergo the force directed layout so it would be possible to limit its application only to the colliding spiders. I'm not planning on adding that to the PR, only leaving it here for future reference.

jvdwetering:
Currently there are no notebooks in the unittests. I'm not sure how you can run those as part of pytest.
[...]

PR #490, opened 2026-07-25
Issue#88 collision

doczenwiry:
## Description
Added support for the detection and highlighting of overlapping nodes. Whenever a node is touched by another, it will be highlighted with a dashed black circle filled in yellow. Furthermore, a box summarising the number of nodes that are overlapping will be drawn at the top-left corner of the figure as long as this number is not zero.

The implementation uses d3-quadtree to query the space of nodes in an efficient way. In order to efficiently detect overlap, all the nodes are assumed to be disks of radius node_size (scaled for BOUNDARY and W_INPUT). The tests include the same circuits as the PR #488 with some overlapping nodes added. A test with an existing circuit including H-boxes has also been added to that same notebook.

The behaviour was tested in Jupyter (running in PyCharm Pro) and JupyterLab (tested on Safari, Firefox & Chrome).

## Related issue
This is 1/3 of Issue #88.

## Possible improvements
The notebook ZXW_demo.ipynb show overlaps detected for W_INPUT and W_OUTPUT nodes even though they do not visibly touch. The overlap condition used falls short for the triangle shape of W_OUTPUT and could be tweaked.

## Checklist
- Tests are included in scratchpads/issue#88.ipynb.
- CHANGELOG.md and documentation (gettingstarted.ipynb) have been updated.
- Comments were added to document the function detect_overlaps(..) added to zx_viewer.inline.js.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Hence, occasionally changes will be backwards incompatible (although they will a
 - Regression test for `unsafe_pivot` with boolean pivot phases (by @dlyongemallo).
 - Added support for zooming in D3 drawings (by @doczenwiry).
 - `apply_to_boolean_axels` opt-in flag to `(unsafe_)pauli_push`, so boolean parameters are not transformed into non-boolean phase by default (by @dlyongemallo).
+- Added support for collision detection in D3 drawings (by @doczenwiry).
 
 ### Changed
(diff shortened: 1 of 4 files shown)


jvdwetering:
@claude review this

jvdwetering:
@doczenwiry Thanks for the new feature! Please consider the feedback above and take it into account. If any of it is non-sensical or not necessary feel free to disregard it.

doczenwiry:
Thanks ! I've updated the code following the feedback. Most of it felt quite sound and pointed out a few quirks I missed. The information to fix the W-node overlap false positives was especially useful ! I can add a few comments perhaps.

- I left the commented out console.log(..) as this is a more complicated piece of code. If anyone attempts to tweak the approximation underlying the overlap detection (i.e. all nodes are assumed to be circles) that'll be helpful.

- I took the liberty to fix a small bug in the function update_hboxes(..) from zx_viewer.inline.js; there was a missing let/var keyword for the variable offset so that pyzx.draw(..) with auto_hbox=True now properly renders. Dragging doesn't seem to work though.

- It might not be the cleanest thing to take those changes and those from PR #488 and put them into zx_viewer.inline.js as part of this PR but I can add them if that's fine by you.

jvdwetering:
Agreed on the zx_viewer.inline.js situation. That is something to be considered separately.
Thanks for the work!

PR #491, opened 2026-07-27
Enforce type annotations in `pyzx.graph` and `pyzx.circuit`

96-LB:
Follow-up to #461 and #473 which fills in the remaining missing type annotations and enables the mypy setting to require type annotations on all functions. The deprecated `graph_ig.py` and `graph_gt.py` are excluded because the classes aren't even possible to instantiate (the constructor raises an exception). Some additional cleanup in this PR:
- The `Scalar` class now raises a `ValueError` if a symbolic scalar is converted to a number (this previously crashed anyway, but now there's a message)
- A couple type annotations for `BaseGraph` from #461 were updated to be more useful
- Type annotation syntax was modernised as much as possible without relying on typing_extensions
- Imports were standardised with isort and types were imported from `collections` instead of `typing` where appropriate
- Minor formatting changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,13 @@ warn_return_any = false
 warn_unused_configs = true
 
 [[tool.mypy.overrides]]
+module = [
+    "pyzx.graph",
+    "pyzx.circuit"
+]
+disallow_untyped_defs = true
+
+[[tool.mypy.overrides]]
 module = "pyzx.optimize"
 ignore_errors = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,8 +76,8 @@ warn_unused_configs = true
 
 [[tool.mypy.overrides]]
 module = [
-    "pyzx.graph",
-    "pyzx.circuit"
+    "pyzx.graph.*",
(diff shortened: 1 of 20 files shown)


jvdwetering:
@claude review this

jvdwetering:
Maybe this is also the signal to formally deprecate graph_ig and graph_gt, since no one uses these and they are not fully implemented anyway.

jvdwetering:
@96-LB Can you confirm that the suggestions by Claude are not necessary for this PR?

jvdwetering:
@claude check whether the new changes have fixed the issues that were identified earlier.

96-LB:
I'll touch up those last couple nits, one second.

96-LB:
Okay, should be good to go now.

Item 3. is out of scope for the PR, and I still opted not to include an `isort` configuration because standardising imports is definitely not a decision that should be made in this PR—I chose to do a default sort here primarily because imports had become a huge mess. However, I do think it would be beneficial to enforce import sorting in a future PR, and I agree with Claude's suggestion to use the black profile.

Issue #131, opened 2023-08-21
Add more ZXW rewrite rules


RazinShaikh:
PyZX only has the W fusion rule for now. PyZX should support some of the common rewrite rules of the ZXW calculus. Here are the few rules I have in mind:

- [ ] ZW bialgebra
- [ ] WW bialgebra (I prefer the [W projector version](https://arxiv.org/pdf/2302.12135.pdf) instead of the fermionic swap version because fermionic swap doesn't generalize to qudits)
- [ ] Trialgebra
- [ ] Pcy - any Z copies through W

It would be nice to support these rules in both forward and backward directions because for rules like trialgebra, we often want to switch from W (or X) to X (or W), apply an X (or W) rewrite, and then switch back using the opposite direction of trialgebra.

jvdwetering:
It would probably be best to add these into their own file (something like wrules.py), just like how the H-box rules are in their own file.

doczenwiry:
Hello everyone,

I'm looking into this issue and I have some questions about the ZW-bialgebra rule. Just a brief summary to make sure I'm understanding correctly; a W-node is represented in PyZX by connecting together a pair of vertices; one W_INPUT and one W_OUTPUT. From there, a single other vertex is attached to the W_INPUT while the W_OUTPUT can have any number of other vertices attached to it (I assume there is a general form beyond the complete (2,2)-bipartite pattern).

I've expanded the BZW-rule from [https://arxiv.org/abs/2302.12135] into the pattern below to more easily visualise the representation in PyZX. A triangle represents a W_OUTPUT vertex while a small black circle represents a W_INPUT vertex (I_1, I_2, O_1 and O_2 are placeholders).

<img width="512" height="384" alt="Image" src="https://github.com/user-attachments/assets/d71f28d3-4535-4408-8ae8-8b156c2078f5" />

My questions are:
- Can I assume that every W_OUTPUT in the graph has exactly one W_INPUT attached to them with a SIMPLE edge ?
- Are there any restrictions on the type of the vertex attached to W_OUTPUT through W_INPUT (i.e. I_1, I_2) ?
- Are there any restrictions on the type of the other vertices attached to the Z-spiders (i.e. O_1, O_2) ?

RazinShaikh:
Thanks for the interest in this issue. 

1. W_INPUT is connected to only one W_OUTPUT and vice versa. They are connected via a special edge type called W_IO, instead of the simple edge 

[...]

PR #492, opened 2026-07-28
Issue 131 (ZW-bialgebra rule)

doczenwiry:
## Description
I've added support for the ZW-bialgebra rule. The forward and reverse directions are implemented and integrated through pyzx.simplify.zw_bialgebra_simp and pyzx.simplify.zw_bialgebra_simp_op.

## Raised problem
- When testing the equivalence, I noticed that the compare_tensors(..) takes a very long time when using 5 qubits (~20 seconds on a MacMini M2). I assume this has something to do with the presence of W-nodes. I've included the steps to recreate the issue at the bottom of the notebook scratchpads/issue131-bzw.ipynb.

## Related issue
This is part of issue #131 

## Checklist
- I have added a batch of unit tests under tests/test_wrules_bzw.py. The equivalence was tested with preserve_scalar=True.
- The CHANGELOG.md and documentation (rules.ipynb) are updated.
- All the functions I added have a docstring.


jvdwetering:
@claude review this

jvdwetering:
@doczenwiry Thanks for the new feature! Please consider the feedback above and take it into account. If any of it is non-sensical or not necessary feel free to disregard it.

jvdwetering:
Hi @doczenwiry it looks like Claude flagged a number of relevant issues (some typos and potential edge cases). Can you confirm that these are not issues, or can you fix them?

doczenwiry:
Hi @jvdwetering, indeed; they are issues and I'll take care of them this week :)

doczenwiry:
I have already addressed several items on the list of the review. However, many of my tests/ are failing since the last version of master I used and I'll have to fix them. I will get back to you during the week when it is done.

doczenwiry:
I have adressed all the issues raised by Claude but I want to add a couple of tests for the cases that it pointed out.

I still have trouble fixing my existing tests, though. The ones that crash are calling compare_tensors(..) to determine equivalence or non-equivalence based on the rewrites that are attempted. Diving deeper into the error brought me to the method tensorfy_rw_subtree(..) that complains about "ValueError: Invalid vertex type: 4". It seems that method only accepts vertices of type Z or BOUNDARY and considers W_INPUT as invalid. Is there another way I can check equivalence of ZX-graphs involving W-nodes ?

jvdwetering:
Ah, this might be due to the newly merged #495. It checked whether there were any H-boxes and went to the default otherwise, but it didn't check for the presence of W-spiders.
For now you can change your tests to tell it to use the naive tensor contraction strategy instead of the automatic one.

doczenwiry:
That was very helpful; all my existing tests are working again :)
I'll just add a few more especially around the types of edges and self-loops connected through the W_INPUT and Z vertices (that was an issue raised by Claude I hadn't be careful about).
The PR should be ready on Monday.

doczenwiry:
Finally wrapped it up :) I've added a few tests for cases where the type of external edges (towards inputs or outputs) are of Hadamard types and when an input is also an output. It would be a good idea to have someone who regularly uses the ZXW-calculus to take a look at my tests to make sure they cover enough as I'm not familiar with the breadth of cases out there.

pyzx/rewrite_rules/wrules_bzw.py
RazinShaikh, line 1: Hi @doczenwiry, I am wondering if the Z-W bialgebra needs to be implemented in a separate file. Currently the bialgebra rule file takes care of bialgebra between Z & X as well as X & H and I believe that it can also easily integrate with the Z & W bialgebra. The only tricky bit with W is taking care of the W-in and W-out nodes but other than that, the logic of the bialgebra is identical. To avoid code duplication, I would recommend integrating your implementation with the already existing bilagebra rule file.

RazinShaikh (changes requested):

pyzx/rewrite_rules/wrules_bzw.py
doczenwiry: Hi @RazinShaikh . Thank you for the feedback. I originally followed the recommendation in the first comment following the description of #131 but I didn't check how the H-box rules overlap with the rest of rewrite_rules. I can integrate the contents of wrules_bzw.py into bialgebra.py if you believe it is relevant. However, I have other projects that are keeping me busy so I'd need some time to get around to doing this.

jvdwetering:
I haven't looked in detail at how much overlap there is, but I do feel like W-spiders are different enough (as they consist of two nodes), that the increased complexity might make it worthwhile to keep it split up. After all, we made all the rewrites separate files because the files were becoming too long and complicated.

RazinShaikh:
That's a reasonable point to not overcomplicate one file. I suppose that ideally, we could have 3 files: `bialgebra_common.py`, `bialgebra_rule.py`, and `zw_bialgebra_rule.py`.

The bialgebra_common.py can have methods like `match_complete_bipartite(graph, vertices1, vertices3)`, `create_complete_bipartite(...)`, etc.

But it's not strictly necessary to make such a separation right now. I can see one future benefit of such factorization. In ZXLive issue https://github.com/zxcalc/zxlive/issues/376 we would like to add support for custom generators and let them satisfy rules like "bialgebra with Z nodes". If we had the core of bialgebra in `bialgebra_common.py`, we could easily add this feature (once the generator has been implemented)

pyzx/rewrite_rules/wrules_bzw.py
RazinShaikh: Thanks for the reply. Please see the messages on the PR.

RazinShaikh:
But more importantly, we should consider how the api is exposed externally. If you call the bialgebra rule (or press the bialgebra button in ZXLive), it will apply Z-X or X-H rule but not Z-W rule. And we have a separate button for Z-W bialgebra but not for X-H bialgebra. That's an inconsistent choice and we should at least make that consistent.

jvdwetering:
That's a good point that it should be doing the same thing. That could also be done in an easy way with a custom matcher that just tries to match both types and applies whichever one works (since they can't both work).

Issue #415, opened 2026-03-06
Automatically select strategy for tensorfy


RazinShaikh:
The tensorfy method explains what kind of graphs different strategies are good for but the default strategy is 'naive'. It would be useful to have a default 'auto' option that automatically detects if the graph is sparse/dense and pick a strategy.
```python
def tensorfy(g: 'BaseGraph[VT,ET]',
             preserve_scalar: bool = True,
             strategy: str = 'naive',
             verbose: bool = False) -> NDArray[np.complex128]:
    """
    Returns a multidimensional numpy array representing the linear map the ZX diagram implements.
    Available simulation strategies are:

    - 'naive': good for sparse graphs
    - 'rw-greedy-b2t': rank-width with greedy bottom-to-top heuristic
    - 'rw-greedy-linear': rank-width with greedy-linear heuristic
    - 'rw-auto': choose the best of 'rw-greedy-b2t' and 'rw-greedy-linear'
    """
```

gauthamkanagaraj:
Hi @RazinShaikh , @jvdwetering , I'm fairly new to PyZX and this issue looked like something I could attempt as a first contribution. I've prototyped an approach for `strategy='auto'` and wanted to check if it's a sensible direction before opening a PR.

I tried predicting what naive would actually cost, rather than classifying the graph up front. The size of its largest intermediate tensor depends only on the diagram's topology (the leg counts at each step) and not on the phases, so it can be computed with a cheap O(V+E) dry run that builds no tensors. In my tests the prediction matched the real `tensorfy` peak on every graph small enough to actually run.

Since the estimate is nearly free, I run it twice, once on the raw diagram and once on its `full_reduce` form, and use `naive` on whichever has the smaller predicted peak, as long as it fits a memory budget. The second estimate seems to matter because `full_reduce` cuts both ways: it shrinks tangled/closed diagrams a lot, but it seems to make circuits worse for `naive` as the graph-like form is more connected, so circuits stay on the raw graph:

| graph | raw peak (log2 entries) | reduced peak (log2 entries) | `auto` picks |
|---|---|---|---|
| 8-qubit circuit | 17 | 36 | naive (raw) |
| dense, 14 spiders | 23 | 8 | naive (reduced) |
| dense, 22 spiders | 48 | 20 | naive (reduced) |

[...]

PR #495, opened 2026-08-06
Default tensorfy strategy

kuyanov:
## Description
Changed the default tensorfy strategy from 'naive' to 'rw-auto'. Since the latter does not yet support H-boxes, manually specified `strategy='naive'` in those test cases.

## Related issue
#415 

## Checklist
- [ ] I have added/updated tests (for bugfixes, please include a regression test, for new features, include new tests either to an existing test file or a new file).
- [ ] For new features: I have updated the documentation.
- [ ] All the functions I added have a docstring parsable by sphinx (for a good example see `pyzx.extract.extract_circuit`).
- [ ] I have added an entry to CHANGELOG.md describing my change.


kuyanov:
All tests are passing except for the multigraph ones. For those, the problem seems to be with `full_reduce`: it leaves duplicate edges in the reduced diagram, but the recursive contraction subroutine assumes there are no such edges. I guess a separate issue should be added.

jvdwetering:
Wouldn't a safer option then be that the automatic strategy detects whether it is a multigraph and then falls back to the naive one?

kuyanov:
The default is now 'auto', which calls 'naive' if either g contains an H-box or it is a multigraph. All tests are passing.

jvdwetering:
Thanks!

PR #513, opened 2026-09-14
Time dependency

JennyLiu0716:
## Description

Adds optional 1+1D (1D space + 1D time) time dependency for circuits: an integer `timestep` and `delay` annotation on ZX-diagram spiders, derived from a circuit's gates and their durations.

`Circuit.to_graph` / `circuit_to_graph` gain an optional `gate_durations` argument — a mapping from a gate class or name to a non-negative integer duration. When given, the gates are scheduled as-soon-as-possible in discrete integer time (`pyzx.circuit.scheduling.schedule_gates`) and every spider of the resulting graph is tagged with `timestep`/`delay` vertex data (`pyzx.graph.time`). With `gate_durations=None` (the default) the output is unchanged.

This enables two things:

- **Space-time cost metrics of a circuit as a ZX-diagram** — `spacetime_metrics(g)` returns the number of timesteps, the time depth, the space-time volume, and the total delay; `time_extent(g)` returns the first/last occupied tick.
- **An induced sub-diagram for a window of timesteps** — `time_slice(g, t)` / `time_slice(g, t0, t1)` returns the sub-diagram of the spiders whose interval meets the window, with cut edges closed off by fresh boundary vertices ordered by qubit — e.g. to exhibit that a sub-circuit implements a phase gadget.

`draw` / `draw_matplotlib` gain a `show_time` option: each timed spider is labelled with its `timestep`, and a wire whose earlier spider has `delay > 0` is drawn as a squiggle labelled with the duration; `delay == 0` (the default) draws exactly as before.

The annotation is stored as ordinary per-vertex `vdata`, so a timed graph is a plain `GraphS`/`Multigraph` — every existing rewrite rule, `extract_circuit`, and `to_tensor` work on it unmodified, with no changes to `gates.py` and no new graph backend. Two notebooks (`doc/notebooks/time_dependency.ipynb`, `doc/notebooks/time_scheduling.ipynb`) walk through the API, including that compatibility check.


JennyLiu0716:
Hi John, thanks for the feedback. 

This is the first piece of the space-time ZX framework from my MPhil thesis (Cambridge, June 2026, supervised by Lia Yeh), which we are now writing up as a paper. We use it to reason about scheduled QEC circuits as ZX-diagrams: the timestep/delay annotation and time_slice let us cut a scheduled circuit at a time window, simplify the induced sub-diagram with PyZX's existing rewriting, and check what map that window implements (e.g. that a subcircuit is a phase gadget, or that a partial protocol prepares the expected intermediate state — used for a fold-transversal surface-code Hadamard and a d=3 magic-state-cultivation circuit in the thesis). The space-time metrics compare implementations of the same logical operation that differ in parallelism.

The reason we would like to put it in PyZX is that the feature is small (two modules, opt-in via one keyword, no change to GraphS or to the output when unused) and it is only useful because it sits next to PyZX's simplification, extraction and tensor checks — a fork would have to track upstream just to keep those working. The later thesis components are deliberately kept out: they are QEC-specific and only depend on PyZX through its public API, so they will live in a separate package on top of PyZX.

> Hi Chenxi, thanks for all the work. A couple of things before I'll take a look at this:
> 
> * You currently add a file to scratchpads, but not to demos. If it makes sense I would promote the .py file to a notebook that explains or shows the features. The scratchpads is really something that I should be deprecating as it was just a workplace for Aleks and me to try out some stuff. The notebook names 'time_dependency' and 'time_scheduling' also don't make it very clear what the difference is between the two. It also looks like you didn't add the new methods into the full API documentation in the docs.
> * You currently add files to the circuit submodule and to the graph submodule. I'm wondering whether it makes sense to make a separate submodule where both files are included. This separates out this code a bit more, and makes it possible to extend it in the future with additional methods. I'm trying to limit the current codebase complexity as much as possible, especially when it comes to feautures that might not be used by the majority of users.
> * It looks like currently none of the code references a paper or something that uses this. Can you explain what you use this for? Why should it be part of the current pyzx codebase, and not exist as a separate fork?

lia-approves:
Thanks John for taking a look. Essentially, the main issue this PR addresses is that if you input a circuit (which has a known depth) into PyZX, from that point on anything about circuit depth is forgotten. Before, you could only talk about ZX diagrams which are quantum computations in terms of graph complexity (vertices and edges), but not in terms of qubit count and depth or runtime. So this lets you optionally label each spider with a timestep.

This is probably best to be in PyZX rather than separate, because the only way to retain time information of an input circuit (say, QASM) is in PyZX's parser, which is where this time information is lost in the first place.

